PR #3, opened 2022-12-19
Form for changing signers/threshold

Zett98:
### Description 
Add the form to have the ability to change the signers of the wallet/signers threshold

#### Depends
- [x] 2

### DIFF
--- a/components/signersForm.tsx
+++ b/components/signersForm.tsx
@@ -1,0 +1,302 @@
+import { validateAddress } from "@taquito/utils";
+import {
+    ErrorMessage,
+    Field,
+    FieldArray,
+    Form,
+    Formik,
+    FormikErrors,
+} from "formik";
+import { FC, useContext, useState } from "react";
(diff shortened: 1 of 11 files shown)


components/signersForm.tsx
```diff
@@ -0,0 +1,295 @@
...
+            ops.push({ add_signers: [...added.values()] })
+        }
+        if (removed.size > 0) {
+            ops.push({ remove_signers: [...removed.values()] })
+        }
+        if (state.contracts[props.address].threshold === requiredSignatures) {
+            ops.push({ adjust_threshold: requiredSignatures })
+        }
+        let params = cc.methods
+            .create_proposal(ops)
```
rueshyna, line 76: it seems something wrong here. the updated setting didn't show in params.  A warning from beacon shows "Transaction is likely to fail"

rueshyna: it works when using new version wallet.

PR #21, opened 2023-02-16
[Refactor] Create wallet UI

JulesGuesnon:
# Description
This PR mainly refactor the wallet creation UI, and includes some fixes as well. 

The main fix is the _gray fix_. Basically, for a reason that I don't know, the `gray` has been overridden in the tailwind config file, which causes all the gray colors to not work. As it can lead to unwanted behaviors, I removed the override value and replaced `gray` by `zinc` in classes, which is way more close of the colors that we use. 

The main changes are in files under `components/create/...`, and the rest is noises due to the fact that a lot of borders came back when fixing the gray color. I'm really sorry about it, this is the last time I'll have noise

https://user-images.githubusercontent.com/36401048/219331435-502012f3-d211-4c03-8cf2-9a5cb7ef33c3.mov

### DIFF
--- a/components/contractLoader.tsx
+++ b/components/contractLoader.tsx
@@ -9,7 +9,7 @@ const ContractLoader: FC<{
       <div role="status" className="mx-auto">
         <svg
           aria-hidden="true"
-          className="text-gray-200 dark:text-gray-600 mr-2 h-8 w-8 animate-spin fill-red-500"
+          className="mr-2 h-8 w-8 animate-spin fill-red-500 text-zinc-200 dark:text-zinc-600"
           viewBox="0 0 100 101"
           fill="none"
           xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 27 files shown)


JulesGuesnon:
Also forgot to mention it:
- the _gray fix_ allows to remove a lot of useless code that was here before
- Even if there's some small style issue here and there, the majority will be removed when I'll implement the new UI with the sidebar which is currently wip

rueshyna (approved):
the change looks good to me :100: . There are few changes we can make for UI of wallet creation in another MR:
- On the second step of creation, 
  - it looks a bit broken on mobile screen
  - I think it'd look nice that when click "add owner", the texts "Owner Name" and "Owner Address" don't repeat again.
- for the "proposal duration" fields, I saw there is error handling but it doesn't seem to work properly. when enter string in alphabet, the whole page will be broken.

PR #22, opened 2023-02-20
[Feature] Wallet management

JulesGuesnon:
# Description
This PR implements a new UI for the wallet. A sidebar on the left, and most of the actions are splitted in individual pages


pages/proposals.tsx
```diff
@@ -0,0 +1,148 @@
...
+      ...proposals.filter(
+        ([_, proposal]) => "Proposing" === proposal.ui.status
+      ),
+    ],
+    [proposals]
+  );
+
+  const currentContract = state.currentContract ?? "";
+  return (
+    <div className="min-h-content relative flex grow flex-col">
```
rueshyna, line 80: I think something wrong with CSS. The popped up Modal isn't in the center of screen. It's in the center of main part. It isn't easy to use when the high of screen is short.

pages/_app.tsx
```diff
@@ -58,16 +67,33 @@ export default function App({ Component, pageProps }: AppProps) {
...
+              <span className="text-xs">Open sidebar</span>
+              <ArrowRightIcon className="h-4 w-4" />
+            </button>
+            {path === "/" && !!state.address && !!state.currentContract ? (
+              <Proposals />
+            ) : (
+              <Component {...pageProps} />
+            )}
           </div>
+          <Footer />
```
rueshyna, line 96: I think it will look nice if the layout of `Footer` align with `Proposals` or `Component`. It looks a bit weird that the texts of footer isn't in the middle.

pages/history.tsx
```diff
@@ -0,0 +1,222 @@
...
+      </div>
+      <main className="h-full min-h-fit grow">
+        <div className="mx-auto h-full min-h-full max-w-7xl py-6 px-4 sm:px-6 lg:px-8">
+          {invalid ? (
+            <div className="mx-auto flex w-full items-center justify-center bg-graybg p-2 shadow">
+              <p className="mx-auto text-xl font-bold text-gray-800">
+                Invalid contract address: {state.currentContract}
+              </p>
+            </div>
+          ) : isLoading ? (
```
rueshyna, line 173: Currently, it looks like if the `state.currentContract` is null, the `invalid` is null so the page will keep loading. so as the other pages.

pages/wallets/[wallet].tsx
```diff
@@ -111,6 +111,7 @@ function Wallet(props: { address: string }) {
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [router]);
+
```
Comment (rueshyna) [inline pages/wallets/[wallet].tsx:114]: Maybe remove this page. This page that us have a quick link to each wallet
- `/wallets/ <multisig wallet address>`

In new design, we should also keep having routes. for example
-  `/<multisig wallet address>/proposals`
- `/<multisig wallet address>/history`
-  etc ...

components/ProposalCard.tsx
```diff
@@ -0,0 +1,186 @@
...
+                type="button"
+                className={
+                  "mx-none mx-auto block w-full self-center justify-self-end rounded bg-primary p-1.5 font-medium text-white hover:bg-red-500 hover:outline-none focus:bg-red-500 md:mx-auto md:w-1/3 md:self-end"
+                }
+                onClick={async e => {
+                  e.preventDefault();
+                  setCloseModal(false);
+                }}
+              >
+                Reject
```
rueshyna, line 135: This can make in other PR. I think the "Reject" should be replace by "Resolve" when number of signature is greater than or equal to threshold.

rueshyna (commented):
The new design UI style is nice :)

pages/wallets/[wallet].tsx
Comment (rueshyna) [reply pages/wallets/[wallet].tsx:114]: Note, currently, after finish creating/importing a wallet, there is a button to go  `/wallets/ <multisig wallet address>`.

rueshyna:
would be nice to have remove button for imported wallets.

pages/wallets/[wallet].tsx
Comment (JulesGuesnon) [reply pages/wallets/[wallet].tsx:114]: Hmm it's a good idea and it's not that complicated, but it would require some refactor to achieve it, so let's keep it for another PR

components/ProposalCard.tsx
JulesGuesnon: Yeah you're right, I'll make it in another PR I think, there's already a lot of changes

pages/history.tsx
rueshyna: image
It looks a bit weird but let's keep it for now. I'm thinking the other idea, let's discuss offline.

rueshyna: image
Can you change it "please import/create a wallet" for now?

(I'm thinking the other idea, let's discuss offline.)

JulesGuesnon: I added the redirection when I don't detect anymore wallets or user is unauthanticated

Zett98:
seeing lots of render errors from slotui component, otherwise LGTM

Zett98 (approved):

PR #24, opened 2023-02-22
[Fix] Nav responsive

JulesGuesnon:
# Description
I'm not sure if I'm the one who broke the nav bar responsive or it was before, but anyway, here is a fix: 

https://user-images.githubusercontent.com/36401048/220556742-ea1366d2-9ecb-4ce5-9ee1-b1aa62d10566.mov


rueshyna (approved):
LGTM. Thanks for the fix, any way :)

PR #26, opened 2023-02-28
[Refactor] Wallet import

JulesGuesnon:
# Description
This PR refactor the UI so it's matching the wallet creation, and it fixes aliases bugs

### DIFF
--- a/components/create/aliases.tsx
+++ b/components/create/aliases.tsx
@@ -120,11 +120,14 @@ function Aliases() {
                     values.validators.map((validator, index) => {
                       return (
                         <div
-                          className="md:p-none flex min-w-full flex-col items-start justify-start space-x-4 p-2 md:flex-row md:rounded-none md:border-none"
+                          className="md:p-none flex min-w-full flex-col items-start justify-start space-x-4 px-2 md:flex-row md:rounded-none md:border-none"
                           key={index}
                         >
-                          <div className="grid grid-flow-col grid-cols-1 grid-rows-3">
-                            <label className="text-white">Owner Name</label>
+                          <div className="grid grid-flow-col grid-cols-1 grid-rows-2">
+                            <label className="text-white">
+                              {index === 0 ? "Owner Name" : ""}
+                            </label>
+
                             <TextInputWithCompletion
                               byAddrToo={false}
                               filter={() => true}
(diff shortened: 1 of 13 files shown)


components/signersForm.tsx
```diff
@@ -238,16 +238,18 @@ const SignersForm: FC<{
...
+                <div className="min-w-full space-y-6">
                   {values.validators.length > 0 &&
                     values.validators.map((validator, index) => {
                       return (
                         <div
-                          className="md:p-none flex min-w-full flex-col items-start justify-start space-y-2 p-2 md:flex-row md:space-y-0 md:space-x-4 md:rounded-none md:border-none"
+                          className="md:p-none flex min-w-full flex-col items-start justify-start space-x-4 px-2 md:flex-row md:rounded-none md:border-none"
                           key={index}
                         >
                           <div className="flex w-full flex-col md:w-auto">
```
rueshyna, line 249: the position of remove button will change when there is an error message in either "Owner name" or "Owner address"

rueshyna (commented):
the rest looks good to me

components/create/aliases.tsx
```diff
@@ -120,11 +120,14 @@ function Aliases() {
                     values.validators.map((validator, index) => {
                       return (
                         <div
-                          className="md:p-none flex min-w-full flex-col items-start justify-start space-x-4 p-2 md:flex-row md:rounded-none md:border-none"
+                          className="md:p-none flex min-w-full flex-col items-start justify-start space-x-4 px-2 md:flex-row md:rounded-none md:border-none"
```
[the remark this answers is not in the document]
rueshyna: It looks like this now. maybe have some space between "index0" and "index1"?
image

when there is an error, the "index1" move down a little bit. when error is gone, the "index1" move back. is it possible to fix it? the same question for signersForm.tsx
image

context/state.ts
```diff
@@ -191,33 +191,30 @@ function reducer(state: tezosState, action: action): tezosState {
       };
     }
     case "removeContract": {
-      let { [action.address]: _, ...contracts } = state.contracts;
```
rueshyna, line 194: the bug seems still there (when imported 3 wallets?)

JulesGuesnon: The bug is FINALLY fixed. The way I fix it is not optimal, but on the other hand, disabling the delete for 3s is a good thing because it'll avoid missclicking and deleting a wallet that you don't want to delete

rueshyna:
LGTM. but don't forget to put "Proposal duration" in import wallet.  (This can be in other PR.)

rueshyna (approved):

JulesGuesnon:
@rueshyna Just added the proposal duration in the import so everything should be good

rueshyna:
Nice! LGTM!

PR #27, opened 2023-02-28
setup CI tag creation compatible auto-deploy

Laucans:
This is a CICD workflow compatible with auto-deploy
Instead of [marigold-dev ](https://github.com/marigold-dev/Multisig-ui/pull/25)

Todo before merge : 
- Create a branch `candidate`
- Create a branch `release`
- Ensure permissions are correct on Github deploy token Url should be https://github.com/marigold-dev/Multisig-ui/settings/actions
image

Workflow : 
- When something is push on candidate branch 
Pre-release tag is created with the format : `vx.y.z-rc.b`. rc = release-candidate, b = version of pre-release
The cicd will detect pattern and chose the last tag containing rc and deploy it on staging. (This part has not been done yet, but I'll finish tomorrow) 

- When you are happy with your release-candidate, merge it into release branch through a PR

- On every other branch or PR the project gonna be built.

@cyrilevos is working on a solution for notifications on slack.

If you see an issue on the workflow please contact me or infra team channel


.github/workflows/docker.yml
```diff
@@ -3,42 +3,72 @@ name: Docker CI
 on:
   push:
     branches:
-      - "main"
+      - '*'
```
rueshyna, line 6: I'm wondering that why we don't run job on specific branches? The other branch seems to build for nothing in this case.

Laucans: It'll be trigger on every branch, but jobs are executed following the name of the branch. 
It's mainly to trigger the build without push every where and check your project can compile.  Because it's a public repo, ci time is free so no trouble here. In case of push on release for example, cache should avoid double build
 
  ```
   - name: Build
        if: ${{ startsWith(github.ref,'refs/heads/') }}
        uses: docker/build-push-action@v4
        with:
          file: ./Dockerfile
          push: false
          cache-from: type=gha
          cache-to: type=gha,mode=max
          build-args: ENV=prod
```

cyrilevos:
LGTM :) Looks to be a nice workflow!

.github/workflows/docker.yml
rueshyna: It's just a little bit my environmental concern for building all branch. It could save some energy to protect our earth. I'd tend to build for all PRs and important branches. but it's unlikely that Multisig will have a lot of active bunch at the same time.

rueshyna:
LGTM

@Zett98 could you help to setup the workflow permission for release branch as description show?

rueshyna (approved):

PR #28, opened 2023-03-01
[Feat] Small UI improvements

JulesGuesnon:
# Description
- Contact in footer
- Icons in sidebar
- Navbar width


components/footer.tsx
```diff
@@ -8,13 +8,35 @@ const Footer = ({
       } lg:text-left`}
     >
       <div className="flex flex-col p-4 text-center text-white">
-        <span className=" text-white">
```
rueshyna, line 11: I think it's a bit weird that copyright, contact, power by listing in column. Would it be better if it look like the footer of this github page

components/Sidebar.tsx
```diff
@@ -244,35 +246,215 @@ const Sidebar = ({
           className={linkClass(path === "/proposals")}
           onClick={onClose}
         >
-          Proposals
```
rueshyna, line 247: would it be nice if the following items, proposals, create a proposal, etc, move slightly to right a little?

rueshyna (commented):
Just some picky comments, it looks good to me in general

rueshyna (approved):
LGTM

PR #29, opened 2023-03-02
rename terms on UI

rueshyna:
- create new wallet -> new wallet
- create a proposal -> new proposal
- top up wallet -> fund wallet
- import existing wallet -> import wallet
- delete -> deleting in setting if canDelete is false


JulesGuesnon (commented):
LGTM

PR #34, opened 2023-03-07
[Fix] Favorite and NaN

JulesGuesnon:
# Description
This PR fixes the favorite contract and the proposal durations inputs:
- Now if you refresh you should have the same contract selected than before
- If you enter a wrong input you should have an error message

### DIFF
--- a/components/create/aliases.tsx
+++ b/components/create/aliases.tsx
@@ -40,7 +40,7 @@ function Aliases() {
     return null;
   }
   const renderError = (message: string) => {
-    return <p className="italic text-red-600">{message}</p>;
+    return <p className="mt-2 italic text-red-600">{message}</p>;
   };
   const initialProps: {
     validators: { name: string; address: string }[];
@@ -62,6 +62,7 @@ function Aliases() {
         const errors: {
           validators: { address: string; name: string }[];
           validatorsError?: string;
+          effectivePeriod?: string;
         } = { validators: [] };
         let dedup = new Set();
         let dedupName = new Set();
@@ -91,12 +92,18 @@ function Aliases() {
           }
           return err;
         });
+
+        if (isNaN(parseInt(values.effectivePeriod as any))) {
+          errors.effectivePeriod = "Invalid duration";
+        }
+
         if (
           result.every(x => x.address === "" && x.name === "") &&
           typeof errors.validatorsError == "undefined"
         ) {
-          return;
+          return errors;
         }
+
         errors.validators = result;
         return errors;
       }}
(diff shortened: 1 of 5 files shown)


components/signersForm.tsx
```diff
@@ -191,12 +192,17 @@ const SignersForm: FC<{
         if (values.requiredSignatures > values.validators.length) {
           errors.requiredSignatures = `threshold too high. required number of signatures: ${values.requiredSignatures}, total amount of signers: ${values.validators.length}`;
         }
+
+        if (isNaN(parseInt(values.effectivePeriod as any))) {
```
rueshyna, line 196: It seems to be a JS thing that `parseInt("12a0test")` will be `12` ?
Maybe use another way to check number? 
(I personally recommend not use `any` if possible)

rueshyna (commented):
It seems that some buttons can't be clicked in this branch
- save button from setting
- continue button of second step from new/import wallet
The above are just the one I checked.

JulesGuesnon:
It should be fixed now

components/create/aliases.tsx
```diff
@@ -91,12 +92,19 @@ function Aliases() {
           }
           return err;
         });
+
+        if (isNaN(Number(values.effectivePeriod))) {
```
rueshyna, line 96: I forgot to say that would be nice to check if number `>0`.

components/signersForm.tsx
```diff
@@ -191,6 +192,12 @@ const SignersForm: FC<{
         if (values.requiredSignatures > values.validators.length) {
           errors.requiredSignatures = `threshold too high. required number of signatures: ${values.requiredSignatures}, total amount of signers: ${values.validators.length}`;
         }
+
+        if (isNaN(Number(values.effectivePeriod))) {
```
rueshyna, line 196: also here

rueshyna (commented):
continue button of second step from import wallet can't click, the others looks good.

components/create/aliases.tsx
```diff
@@ -93,7 +93,8 @@ function Aliases() {
           return err;
         });
 
-        if (isNaN(Number(values.effectivePeriod))) {
+        const parsedNumber = Number(values.effectivePeriod);
+        if (isNaN(parsedNumber) || parsedNumber <= 0) {
```
rueshyna, line 97: Sorry, not equal `0`. it doesn't make sense to have 0 duration and contract part will throw error for 0 duration.

components/signersForm.tsx
```diff
@@ -193,7 +193,8 @@ const SignersForm: FC<{
           errors.requiredSignatures = `threshold too high. required number of signatures: ${values.requiredSignatures}, total amount of signers: ${values.validators.length}`;
         }
 
-        if (isNaN(Number(values.effectivePeriod))) {
+        const parsedNumber = Number(values.effectivePeriod);
+        if (isNaN(parsedNumber) || parsedNumber <= 0) {
```
rueshyna, line 198: the same

rueshyna: my bad. please ignore it.

rueshyna (approved):

PR #35, opened 2023-03-07
[Fix] Limit user actions if not owner

JulesGuesnon:
# Description
This PR disables `create proposal` and `settings` page if the user isn't the owner

### DIFF
--- a/pages/settings.tsx
+++ b/pages/settings.tsx
@@ -1,12 +1,22 @@
+import { useRouter } from "next/router";
 import { useContext, useEffect, useState } from "react";
 import Meta from "../components/meta";
 import SignersForm from "../components/signersForm";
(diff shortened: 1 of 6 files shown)


pages/settings.tsx
```diff
@@ -1,12 +1,22 @@
...
   const state = useContext(AppStateContext)!;
   const dispatch = useContext(AppDispatchContext)!;
   const [canDelete, setCanDelete] = useState(true);
+  const router = useRouter();
+  const isOwner = useIsOwner();
+
+  useEffect(() => {
+    if (isOwner) return;
+
+    router.replace("/proposals");
```
rueshyna, line 18: I think not owner should be allowed to see setting but can't change it, just like imported wallet.

rueshyna (approved):
:100:

PR #38, opened 2023-03-09
unify style

rueshyna:
Unify style of
- spinner: the spinner will all in center
- import wallet page and setting page


components/import/aliases.tsx
```diff
@@ -169,18 +170,24 @@ function Aliases() {
...
-            </label>
-            <Field
-              disabled
-              component="input"
-              name="effectivePeriod"
-              value={formState?.effectivePeriod ?? "Unknown"}
-              className="rounded p-2"
-            />
-          </div>
+          {typeof formState?.effectivePeriod != "undefined" && (
```
JulesGuesnon, line 173: `!!formState?.effectivePeriod` should work and is smaller

components/signersForm.tsx
```diff
@@ -250,6 +250,7 @@ const SignersForm: FC<{
                               {index === 0 ? "Owner Name" : ""}
                             </label>
                             <Field
+                              disabled={props.disabled ? true : null}
```
JulesGuesnon, line 253: Is there a reason for those conditions? Like it looks like it's the same as doing; `if (boolean) {true} else {false}`

JulesGuesnon (commented):
Some really really minor js syntax related stuff, otherwise LGTM!

components/signersForm.tsx
rueshyna: the type is `boolean | undefined`. It looks like not ok to set `disabled = undefined`

JulesGuesnon: Well basically if you don't add the props `disabled` to `<Field />` by default it'll be undefined. So the type of `disabled` is `boolean | undefined` which is matching our `disabled` props. So imo it should work, but this is really some details, and I don't really want to bother you which such a thing as it's really quick, so I'll check myself

JulesGuesnon (commented):
LGTM

JulesGuesnon (approved):

components/signersForm.tsx
rueshyna: Ah, I got it! Thanks!

PR #39, opened 2023-03-09
[Fix] Handling timeout

JulesGuesnon:
# Description
As we still have an issue with taquito confirming if the transactions, I added a timeout so the user can have some feedback if we don't have confirmation. Is set the timeout to `90s` 

image

### DIFF
--- a/components/proposalSignForm.tsx
+++ b/components/proposalSignForm.tsx
@@ -1,6 +1,8 @@
+import { NetworkType } from "@airgap/beacon-sdk";
 import { ErrorMessage, Field, Form, Formik } from "formik";
 import { useRouter } from "next/router";
 import React, { useContext, useState } from "react";
+import { PREFERED_NETWORK } from "../context/config";
 import { AppStateContext } from "../context/state";
 import { version, proposal } from "../types/display";
 import { VersionedApi } from "../versioned/apis";
@@ -59,17 +61,19 @@ function ProposalSignForm({
     return (
       <div className="mx-auto mt-4 w-full text-center text-zinc-400">
         <p>
-          The wallet {"didn't"} confirmed that the transaction has been
-          validated. You can check it in{" "}
+          The wallet {"can't"} confirm that the transaction has been validated.
+          You can check it in{" "}
           <a
             className="text-zinc-200 hover:text-zinc-300"
-            href={`https://ghostnet.tzkt.io/${timeoutAndHash[1]}`}
+            href={`https://${
+              PREFERED_NETWORK === NetworkType.GHOSTNET ? "ghostnet." : ""
+            }tzkt.io/${timeoutAndHash[1]}`}
             target="_blank"
             rel="noreferrer"
           >
             the explorer
           </a>
-          , and if it is, {"it'll"} appears in the history
+          , and if it is, {"it'll"} appear in the history
         </p>
         <div className="w-full space-x-4">
           <button
@@ -96,7 +100,8 @@ function ProposalSignForm({
       <div className="flex w-full flex-col items-center justify-center">
         <ContractLoader loading={loading}></ContractLoader>
         <span className="mt-4 text-zinc-400">
-          Waiting for transaction confirmation (it may takes a few mintues)
+          Sending and waiting for transaction confirmation (It may take a few
+          minutes)
         </span>
       </div>
     );
--- a/components/proposalSignForm.tsx
+++ b/components/proposalSignForm.tsx
@@ -73,7 +73,9 @@ function ProposalSignForm({
           >
             the explorer
           </a>
-          , and if it is, {"it'll"} appear in the history
+          , and if it is, and the proposal is resolved, it will appear in the
+          history, otherwise it will still remain in the proposals with updating
+          status
         </p>
         <div className="w-full space-x-4">
           <button
--- a/components/proposalSignForm.tsx
+++ b/components/proposalSignForm.tsx
@@ -1,6 +1,7 @@
 import { ErrorMessage, Field, Form, Formik } from "formik";
+import { useRouter } from "next/router";
 import React, { useContext, useState } from "react";
-import { AppDispatchContext, AppStateContext } from "../context/state";
+import { AppStateContext } from "../context/state";
 import { version, proposal } from "../types/display";
 import { VersionedApi } from "../versioned/apis";
 import ContractLoader from "./contractLoader";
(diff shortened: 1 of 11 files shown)


components/proposalSignForm.tsx
```diff
@@ -39,17 +44,62 @@ function ProposalSignForm({
...
     );
   }
 
   if (loading && typeof result == "undefined") {
-    return <ContractLoader loading={loading}></ContractLoader>;
+    return (
+      <div className="flex w-full flex-col items-center justify-center">
+        <ContractLoader loading={loading}></ContractLoader>
+        <span className="mt-4 text-zinc-400">
+          Waiting for transaction confirmation (it may takes a few mintues)
```
rueshyna, line 99:
```suggestion
          Sending and waiting for transaction confirmation (It may take a few minutes)
```

components/proposalSignForm.tsx
```diff
@@ -39,17 +44,62 @@ function ProposalSignForm({
...
+          validated. You can check it in{" "}
+          <a
+            className="text-zinc-200 hover:text-zinc-300"
+            href={`https://ghostnet.tzkt.io/${timeoutAndHash[1]}`}
+            target="_blank"
+            rel="noreferrer"
+          >
+            the explorer
+          </a>
+          , and if it is, {"it'll"} appears in the history
```
rueshyna, line 72: if the proposal haven't resolved, users won't see it in history.

components/proposalSignForm.tsx
```diff
@@ -39,17 +44,62 @@ function ProposalSignForm({
...
+        Boolean(resolve)
+      )
+    );
+  }
+
+  if (timeoutAndHash[0]) {
+    return (
+      <div className="mx-auto mt-4 w-full text-center text-zinc-400">
+        <p>
+          The wallet {"didn't"} confirmed that the transaction has been
```
rueshyna, line 62:
```suggestion
          The wallet {"can't"} confirm that the transaction has been
```
would it be better :thinking: ? "didn't" sounds that the wallet even didn't give it a try?  so as the others.

components/proposalSignForm.tsx
```diff
@@ -39,17 +44,62 @@ function ProposalSignForm({
...
+          validated. You can check it in{" "}
+          <a
+            className="text-zinc-200 hover:text-zinc-300"
+            href={`https://ghostnet.tzkt.io/${timeoutAndHash[1]}`}
```
rueshyna, line 66: should check the network type and go the proper network :)

components/signersForm.tsx
```diff
@@ -108,10 +111,60 @@ const SignersForm: FC<{
...
+          validated. You can check it in{" "}
+          <a
+            className="text-zinc-200 hover:text-zinc-300"
+            href={`https://ghostnet.tzkt.io/${timeoutAndHash[1]}`}
+            target="_blank"
+            rel="noreferrer"
+          >
+            the explorer
+          </a>
+          , and if it is, {"it'll"} appears in the proposals
```
rueshyna, line 133:
```suggestion
          , and if it is, {"it'll"} appear in the proposals
```

components/signersForm.tsx
```diff
@@ -108,10 +111,60 @@ const SignersForm: FC<{
...
+    );
   }
+
   if (loading && typeof result == "undefined") {
-    return <ContractLoader loading={loading}></ContractLoader>;
+    return (
+      <div className="flex w-full flex-col items-center justify-center">
+        <ContractLoader loading={loading}></ContractLoader>
+        <span className="mt-4 text-zinc-400">
+          Waiting for transaction confirmation (it may takes a few mintues)
```
rueshyna, line 164:
```suggestion
          Sending and waiting for transaction confirmation (It may take a few minutes)
```

components/transferForm.tsx
```diff
@@ -867,13 +870,66 @@ function TransferForm(
...
+    );
+  }
+
   if (loading && typeof result == "undefined") {
-    return <ContractLoader loading={loading}></ContractLoader>;
+    return (
+      <div className="flex w-full flex-col items-center justify-center">
+        <ContractLoader loading={loading}></ContractLoader>
+        <span className="mt-4 text-zinc-400">
+          Waiting for transaction confirmation (it may takes a few mintues)
```
rueshyna, line 929:
```suggestion
           Sending and waiting for transaction confirmation (It may take a few minutes)
```

context/config.ts
```diff
@@ -10,3 +10,5 @@ export const PREFERED_NETWORK =
   process.env.NEXT_PUBLIC_NETWORK_TYPE === "mainnet"
     ? NetworkType.MAINNET
     : NetworkType.GHOSTNET;
+
+export const DEFAULT_TIMEOUT = 90000;
```
rueshyna, line 14: currently, one block/conformation is about 30s. I think we should be shorter, maybe 45s?

components/transferForm.tsx
```diff
@@ -867,13 +870,66 @@ function TransferForm(
...
+            className="text-zinc-200 hover:text-zinc-300"
+            href={`https://${
+              PREFERED_NETWORK === NetworkType.GHOSTNET ? "ghostnet." : ""
+            }tzkt.io/${timeoutAndHash[1]}`}
+            target="_blank"
+            rel="noreferrer"
+          >
+            the explorer
+          </a>
+          , and if it is, {"it'll"} appears in the proposals
```
rueshyna, line 899:
```suggestion
          , and if it is, {"it'll"} appear in the proposals
```

rueshyna (commented):
In general, it looks good to me :+1:

components/proposalSignForm.tsx
JulesGuesnon: You're right, It sounds better

context/config.ts
JulesGuesnon: Oh didn't know, you're right sounds a bit too long. But I would probably go with 60s to make sure even slow connections have the time to get a confirmation, wdyt?

rueshyna: sounds good.

components/proposalSignForm.tsx
rueshyna: how about something word like this??
```
 and if it is, the resolved proposal will be in the history and the pending proposal will still remain in the proposal with updating status
```

rueshyna (approved):
:+1:

PR #40, opened 2023-03-10
while sign out, the imported wallets should be able show on UI 

rueshyna:
- show sidebar when signing out
- import wallet shouldn't require log in

### DIFF
--- a/pages/import-wallet.tsx
+++ b/pages/import-wallet.tsx
@@ -18,33 +18,7 @@ function Import() {
   const state = useContext(AppStateContext);
   const dispatch = useContext(AppDispatchContext);
 
-  useEffect(() => {
-    const connectWallet = async (): Promise<void> => {
-      try {
-        await state?.beaconWallet!.requestPermissions({
-          network: {
-            type: PREFERED_NETWORK,
-          },
-        });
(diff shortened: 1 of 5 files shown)


pages/import-wallet.tsx
```diff
@@ -18,33 +18,7 @@ function Import() {
...
-        router.replace("/");
-      }
-    };
-    (async () => {
-      if (!state?.address && state?.beaconWallet) {
-        await connectWallet();
-      }
-    })();
-  }, [state, dispatch, router]);
+  useEffect(() => {}, [state, dispatch, router]);
```
JulesGuesnon, line 21: Why everything has been deleted?

JulesGuesnon (commented):
So overall looks good, but while testing I had another deletion issue. The contracts were deleted, but there was still aliases, so I'll try to push a fix in this PR

pages/import-wallet.tsx
rueshyna: In the step of importing wallet, no needed for making Tezos transactions so the login action isn't actually required.

JulesGuesnon:
Actually, I think I just didn't paid attention and thought an alias was a contract but it was a user alias. I fixed small UI padding issue tho

JulesGuesnon (approved):

PR #43, opened 2023-03-15
[Fix] Prevent erasing all the aliases

JulesGuesnon:
# Description
I thought I fixed a bug with the address book, but I actually introduced one when adding a new signer, it would erase all the aliases basically


rueshyna (approved):
Thanks for the fix!

PR #44, opened 2023-03-15
[fix] proposal form

rueshyna:
1. fix default value of lambda form
2. modify some terms
3. modify style

for the change 1 and 3, please see the snapshots:
before:
image
after:
image

### DIFF
--- a/components/transferForm.tsx
+++ b/components/transferForm.tsx
@@ -228,7 +228,7 @@ function RenderItem({
   const { values, setFieldValue, getFieldProps } = useFormikContext<any>();
   if ("select" == item?.type) {
     return (
-      <div className="flex w-full flex-col border-2 border-white p-2">
+      <div className="flex w-full flex-col border-2 p-2">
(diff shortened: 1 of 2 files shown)


JulesGuesnon (commented):
Overall LGTM. I think it would be a good idea to remove all the borders in this PR as we removed them everywhere except here I think. Also adding `rounded` class to buttons and inputs so it's more consistent with the rest

rueshyna:
> I think it would be a good idea to remove all the borders in this PR as we removed them everywhere except here I think. 

I removed the outer border. This form is in nest structuring so it would be unreadable if we remove all the inner border. We need a redesign for this form.
image


> Also adding rounded class to buttons and inputs so it's more consistent with the rest

In this page, the other form isn't rounded so I still keep it and do a minimal change here. If I have a time, I'll try to do it in next PR :smiling_face_with_tear:.

JulesGuesnon (approved):
Ok let's go like that for now

PR #51, opened 2023-03-21
[Fix] Small bugs

JulesGuesnon:
# Description

This PR fixes:
- Homepage wording
- Handles error messages for contract call and lambda parsing
- Show error when mutez amounts are `<= 0`
- Show the exact proposal duration time
- Show an error when importing a non TzSafe contract
- Detecting correctly FA2 transactions

### DIFF
--- a/components/import/basic.tsx
+++ b/components/import/basic.tsx
@@ -10,16 +10,19 @@ import { AppStateContext } from "../../context/state";
 import { signers, toStorage } from "../../versioned/apis";
 import Spinner from "../Spinner";
 
+const renderError = (message: string) => (
+  <p className="mt-1 italic text-red-600">{message}</p>
+);
+
 function Basic() {
   const [isLoading, setIsLoading] = useState(false);
+  const [error, setError] = useState<undefined | string>(undefined);
 
   const { activeStepIndex, setActiveStepIndex, formState, setFormState } =
     useContext(FormContext)!;
   const state = useContext(AppStateContext)!;
-  let params = useSearchParams();
-  const renderError = (message: string) => (
-    <p className="italic text-red-600">{message}</p>
-  );
+  const params = useSearchParams();
+
   let [initialState, set] = useState({
     walletName: "example-wallet",
     walletAddress: formState?.walletAddress || "",
@@ -42,6 +45,8 @@ function Basic() {
       enableReinitialize={true}
       initialValues={initialState}
       validate={async values => {
+        if (!!error) setError(undefined);
+
         let errors: any = {};
         if (validateContractAddress(values.walletAddress) !== 3) {
           errors.walletAddress = `Invalid address ${values.walletAddress}`;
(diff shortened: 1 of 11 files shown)


components/transferForm.tsx
```diff
@@ -493,7 +493,10 @@ function Basic({
         if (!exists) {
           errors.walletAddress = `Contract does not exist at address ${values.walletAddress}`;
         }
-        if (isNaN(Number(values.amount))) {
+
+        const convertedValue = Number(values.amount);
+
+        if (isNaN(convertedValue) || convertedValue <= 0) {
```
rueshyna, line 499: we should also check the amount in "Fund wallet"

components/transferForm.tsx
```diff
@@ -514,7 +517,7 @@ function Basic({
                 <Field
                   name="amount"
                   className=" w-full rounded p-2 text-black"
-                  placeholder="0"
```
rueshyna, line 517: the amount in "Contract Execution" is an exception that can be `0` but not `<0`

components/import/basic.tsx
```diff
@@ -69,32 +74,44 @@ function Basic() {
...
-          ...values,
-          validators,
-          requiredSignatures: storage.threshold.toNumber(),
-          effectivePeriod: storage.effective_period.toNumber(),
-        };
-        setFormState(data as any);
-        setActiveStepIndex(activeStepIndex + 1);
-        setIsLoading(false);
+          if (version === "unknown version") {
+            throw new Error("The contract is not a TzSafe contract");
```
rueshyna, line 88: this error wasn't handled on UI.
import `KT1BMLhaeartDoLXjiQZorD3yKQj6SoH6c5R` to reproduce

components/import/basic.tsx
```diff
@@ -69,32 +74,44 @@ function Basic() {
...
+          }
+
+          let balance = await state?.connection.tz.getBalance(
+            values.walletAddress
+          );
+          let v = toStorage(version, storage, balance);
```
rueshyna, line 94: is it possible to have errors here? if yes, we should handle it.

versioned/interface.ts
```diff
@@ -161,9 +161,9 @@ abstract class Versioned {
               }
               const p = new Parser();
               try {
-                const _ = p.parseScript(x);
+                p.parseScript(x);
               } catch {
-                return "Bad lambda";
+                return "Unable to parse the lambda";
```
rueshyna, line 166: In which case, I'll see this error on UI? It doesn't seem to work on UI.

rueshyna (commented):
although the placeholder of amount is 1, but the default value is still 0.

components/import/basic.tsx
JulesGuesnon: Hum are you sure it's the right contract? I imported it and I got the error saying: `The contract is not a TzSafe contract`

components/import/basic.tsx
JulesGuesnon: Everything is in a try/catch so it's already handled

versioned/interface.ts
JulesGuesnon: You can try with this input `qdfqdf!`, it's working on my side. Also you may need to blur the input to trigger a check

components/import/basic.tsx
rueshyna: It shows up on UI now :thinking: (It was nothing there)

versioned/interface.ts
rueshyna: close the thread, as the new request is to remove lambda execution.

rueshyna (approved):

PR #57, opened 2023-03-26
[Feature] sandbox mode

rueshyna:
prepare sandbox of both tezos-node and tzkt's components
- docker-compose.stateless.yml: when containers is done, all the data will be gone
- docker-compose.stateful.yml: the data will remain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,16 @@ And serve the build with:
 ```bash
 npm start
 ```
+
+## Sandbox
+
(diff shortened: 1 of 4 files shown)


JulesGuesnon (commented):
The docker part looks good, but I think it lacks documentation. It would be nice to have explanations of the purpose and how to use all of this in the readme

rueshyna:
updated readme

README.md
```diff
@@ -30,8 +30,21 @@ Once you updated the code, you can build it with:
 npm run build
 ```
 
-And serve the build with:
+And server the build with:
```
JulesGuesnon, line 33: I think it's a typo, it should be `serve` as it was before

JulesGuesnon (approved):
Only a small typo, but otherwise LGTM

README.md
rueshyna: modified.

PR #60, opened 2023-04-04
[Refactor] Proposal duration

JulesGuesnon:
# Description
This PR changes the UI for proposal duration. Rather than inputing seconds, you can now input days, hours and minutes. You can input only integers, and one of the field must be filled otherwise there should have an error

image


JulesGuesnon:
Also added a quick commit to fix the logo which wasn't working in production

rueshyna:
would it be a nice change to start unit test for proposal duration? :)

rueshyna:
@JulesGuesnon let's do that and will see what else can be added/refined later.

PR #61, opened 2023-04-05
[Fix] Making a nice responsive on every page

JulesGuesnon:
# Description
This PR fixes the responsive on most the pages. I would need someone to test it too, to make sure I didn't miss anything. Also this PR should be merged after #60 is merged, so I can make sure the responsive isn't broken on the new changes


aguillon:
At Jules' request and in order to become more familiar with the project, I've tested the UI a bit. Overall, the UI seems pretty good, with minor tweaks that could be fixed in this PR as well IMO.

I've tested the responsiveness during the creation of a wallet, the creation of a proposal and the confirmation of a proposal. It looks good to me in width. I guess the UI could be more responsive in height too (see following screenshot) but I don't think it matters that much.
responsiveness1

Some error messages could be improved: 
- "already exists" when adding an account that's already listed, I suggest "This account is already participating" or "Please enter each account only once".
- "invalid days" when entering "0" as days: what if we want proposals that last less than 1 day? I suggest "Proposals must be put to vote for at least 1 day"

This PR does not fix the issue with mutez.

I've tested on mobile as well and I think there's a bit of room for improvement (screenshots from Firefox for Android):
- I've arrived on the expanded sidebar view by accident and was a bit confused on how I should reduce the bar. Perhaps the "← Close sidebar" option should be more visible? Screenshot_20230406-174504_Firefox
- In horizontal mode, the UI seems unnecessarily stacked: 
Screenshot_20230406-174612_Firefox
Screenshot_20230406-174626_Firefox

aguillon (changes requested):
No opinion on the code.

Please change the two error messages I listed (quick to fix) as well as the "Close sidebar" message, if you agree.

JulesGuesnon:
- Updated the error messages
- Brighten the text color of `open sidebar` and `close sidebar`
- Fixed the issue of `Invalid days` now 0 is a correct value, but if every values are 0 there's an error

aguillon:
> Fixed the issue of Invalid days now 0 is a correct value, but if every values are 0 there's an error

I think the minimum should be something like 15 minutes or even 1 hour, though. If users set a limit so low that proposals expire almost immediately after being posted, that's an issue as well. I actually can't seem to fix this, probably because my computer is a bit out of sync.

>  Brighten the text color of open sidebar and close sidebar

That's better, thanks. Another issue: when testing on mobile (Firefox for Android), clicking "open sidebar" does nothing, even after associating a wallet. I don't have this issue on Firefox for Linux (even with a small page).

Another issue: when I'm on the "Settings" page and I refresh the page, I'm taken back to the "Proposals" page. Actually it happens with from other pages as well, but there's little reason to refresh other pages I think — whereas Settings can be refreshed when the user wants to see the new settings.

Good luck!

rueshyna:
1. We don't force user to add an alias/address.
image

2. For Navbar,
- I'd say "address book", "new wallet", "import wallet" should be right align.  
- remove the red avatar circle nearby the account, we don't have the circle in desktop version.   
- I'd say it's a little troublesome to click twice to see "sign out" option. would be nice to directly display with  "address book", "new wallet", "import wallet".
- would it be a good idea that the account(tzx) and network inform should should display directly in Navbar so users don't need to have a one click to check.
image

3. close menu from Navbar when open sidebar.
image

4. for ProposalCard,
- Param -> Param/Tokens
- the texts are overlap. would it be easy if we display like this in mobile version?
- [nice to change] I think if yes/no selection replace with checkbox would be easier for user.
```
Content
Function   Transfer
Metadata   -
Amount     - 
Address    KTabc...efghi
Entrypoint abcdefg
Param/Token      "param"

Function  Execute Contract
Metadata   -
Amount     - 
Address    KTabc...efghi
Entrypoint abcdefg
Param/Token      "param"


Function  Execute Contract
Metadata   -
Amount     - 
Address    KTabc...efghi
Entrypoint abcdefg
Param/Token      "param"

Active 
# same for active
```
image
image


5. when no wallet is imported, there is no sidebar to open but it still shows "Open sidebar ->"
image

rueshyna:
> Another issue: when I'm on the "Settings" page and I refresh the page, I'm taken back to the "Proposals" page. Actually it happens with from other pages as well, but there's little reason to refresh other pages I think — whereas Settings can be refreshed when the user wants to see the new settings.

@JulesGuesnon 
If it's easy to make a change here, then please proceed. otherwise, I think it can come with routing feature, which should come in this quarter. 

> I think the minimum should be something like 15 minutes or even 1 hour, though.

I'd like put some warning messages for it but not force user to follow. 
warning message should warn users when
- proposal duration is low.
- user remove itself from owners.

> If users set a limit so low that proposals expire almost immediately after being posted, that's an issue as well. I actually can't seem to fix this, probably because my computer is a bit out of sync.

It does have some problem with sync issue when things happen too quick. @JulesGuesnon  can you check if it's a quick fix? If not, please create an issue and we can prioritize fix later.

JulesGuesnon:
> If it's easy to make a change here, then please proceed. otherwise, I think it can come with routing feature, which should come in this quarter.

I know what's happening, but yeah I'll include a fix with the new routes, now it's a bit off topic I think

I agree with the rest, and I'll include new warnings and check for the sync issue

aguillon:
I don't have much time for reviews this week and I see that @rueshyna is back, so I'll skip reviews from now on. :wave:

rueshyna:
1. when signed out, the text `sign out` should be hide.
image

2. To align with other account presentation, I think the account should display 5 chars in both head and tail.
image

3. the date in active is too closed to account, especially, the hyphen.
image

4. should be Params -> Params/Token everywhere.
image

components/signersForm.tsx
```diff
@@ -552,9 +567,14 @@ const SignersForm: FC<{
...
               {/* @ts-ignore*/}
-              {!!errors.proposalDuration &&
-                // @ts-ignore
-                renderError(errors.proposalDuration)}
+              {!!errors.proposalDuration
+                ? // @ts-ignore
+                  renderError(errors.proposalDuration)
+                : currentDuration < PROPOSAL_DURATION_WARNING
+                ? renderWarning(
+                    "Proposal duration is low, you may not be able to resolve proposals"
```
rueshyna, line 575: when proposal is expired, users still can resolve the proposals. Therefore, I suggest s/resolve/execute.

rueshyna (approved):

pages/address-book.tsx
```diff
@@ -111,130 +108,148 @@ function Home() {
...
-                                          Name
-                                        </span>
-                                      </label>
+              {({ values, errors, setTouched }) => {
+                const hasNoError =
+                  ((
+                    errors.validators as
+                      | { name: string; address: string }[]
+                      | undefined
+                  )?.every(v => v.address === "" && v.name === "") ??
```
rueshyna, line 117: when click "Add an address" button, 
```
Unhandled Runtime Error
TypeError: Cannot read properties of null (reading 'address')

Source
pages/address-book.tsx (117:34) @ address

  115 |     | { name: string; address: string }[]
  116 |     | undefined
> 117 | )?.every(v => v.address === "" && v.name === "") ??
```

components/ProposalCard.tsx
```diff
@@ -127,7 +127,7 @@ export const RenderProposalContent = ({
           token_id,
         }),
       };
-    } else {
+    } else if (metadata.meta) {
```
rueshyna, line 130: In which case, we will have `meta` in `metadata`?

JulesGuesnon: Tbh I don't know. I had a crash with one of my wallets, so I just logged the data and handled the case to avoid crashes. It may be some legacy of the previous contract version

rueshyna (approved):

PR #64, opened 2023-04-19
[Feat] Documentation in sidebar

JulesGuesnon:
# Description

It was mentioned during the demo day that we want to have another documentation link, so here it is in the sidebar:

image


rueshyna (approved):
LGTM

Issue #62, opened 2023-04-06
[UI][ContractExecution]Proposal for a contract execution does not list all the entrypoints


aguillon:
When trying to make a proposal to call an entrypoint of [this contract on ghostnet](https://ghostnet.tzkt.io/KT1T42LtbG7skEACAkWxNJGqdcCn52Nhp3pH/entrypoints), I've noticed that some of them do not appear on the UI:

![contract-entrypoints](https://user-images.githubusercontent.com/7795522/230435052-6833cb59-79bb-4d4d-a97c-f41b0758ee1d.png)

For instance, divestLiquidity is listed by tzkt or `octez-client` but does not appear on the UI.

rueshyna:
Thanks for reporting. 
It's a "nested" entrypoint. The `divestLiquidity` is under `use`. If you select `use`, you'll see more entrypoints.
We handle entrypoint in taquito, which doesn't treat `divestLiquidity` is entrypoint.
I'll make some fix on UI presentation.

aguillon:
Oh, first time I see this, interesting. So it's actually just a minor UI bug, then, since changing the "second" entrypoint twice lets you provide the correct arguments.

PR #66, opened 2023-04-21
[Fix] flatten entrypoints and remove "%default" in lambda

rueshyna:
-  #62 
   -  see `parseContract` for the fix,  flatten entrypoints.
- remove %default entrypoint in lambda, explicit %default entrypoint will cause Tezos failed.

recommend reviewing commit by commit.


JulesGuesnon (approved):
LGTM

PR #70, opened 2023-04-28
[Feature] Delegation

JulesGuesnon:
# Description
This PR adds wallet delegation. To do so we have a field in the settings, and if it's filled, it means you want to delegate, and if you erase it, you'll "undelegate"

### DIFF
--- a/components/signersForm.tsx
+++ b/components/signersForm.tsx
@@ -488,7 +488,7 @@ const SignersForm: FC<{
               bakerAddressRef={bakerAddressRef}
             />
             <div className="mb-2 self-start text-left text-lg text-white">
-              Change wallet participants below
+              Wallet participants
             </div>
             <ErrorMessage name={`validatorsError`} render={renderError} />
             <div className="mb-2 grid w-full grid-flow-row items-start gap-4">
--- a/components/signersForm.tsx
+++ b/components/signersForm.tsx
@@ -145,7 +145,10 @@ const SignersForm: FC<{
     txs: { name: string; address: string }[],
     requiredSignatures: number,
     effectivePeriod: number | undefined,
-    bakerAddress: string | undefined
+    {
+      bakerAddress,
+      oldBakerAddress,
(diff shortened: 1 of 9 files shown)


components/signersForm.tsx
```diff
@@ -306,6 +412,23 @@ const SignersForm: FC<{
           errors.proposalDuration = "One value must at least be more than 0";
         }
 
+        if (!!values.bakerAddress) {
+          if (validateAddress(values.bakerAddress) !== ValidationResult.VALID)
+            errors.bakerAddress = `Invalid address ${values.bakerAddress}`;
+          else {
+            try {
+              const account = await fetch(
+                `https://api.ghostnet.tzkt.io/v1/accounts/${values.bakerAddress}`
```
rueshyna, line 421: using entry of tzkt should be from config

components/signersForm.tsx
```diff
@@ -306,6 +412,23 @@ const SignersForm: FC<{
...
+              ).then(res => res.json());
+
+              if (account.type !== "delegate" || !account.activationLevel)
```
rueshyna, line 424: I forgot one thing if there is a `deactivationLevel`, we should show error message like, baker is inactive.

https://ghostnet.tzkt.io/tz1VSUr8wwNhLAzempoch5d6hLRiTh8Cjcjb/operations/

components/signersForm.tsx
```diff
@@ -576,6 +687,16 @@ const SignersForm: FC<{
                   )
                 : null}
             </div>
+            <div className="mt-4 w-full">
+              <label className="block text-white">Delegate wallet</label>
```
rueshyna, line 691: I am thinking maybe the title should be like  "Change wallet participants below" in line 489 so as the "Threshold" and "Proposal duration". what do you think?

rueshyna (commented):
In proposal card, the function name should be "Set delegate". if it's setting baker, the baker's address should be showed in `Address`.  if it's cancelling, show "-".

rueshyna (approved):

PR #73, opened 2023-05-03
[Feature] Failed proposal keep state

JulesGuesnon:
# Description
This PR allows to keep a proposal state if the proposal creation fails. This way, the user can update what needs to be updated and submit another proposal

### DIFF
--- a/components/Sidebar.tsx
+++ b/components/Sidebar.tsx
@@ -169,7 +169,7 @@ const Sidebar = ({
   return (
     <aside
       className={`fixed left-0 bottom-0 ${
-        state.hasBanner ? "top-32" : "top-20"
+        state.hasBanner ? "top-32" : "top-16"
       } z-10 w-72 bg-zinc-700 px-4 py-4 md:py-8 ${
         isOpen ? "translate-x-0" : "-translate-x-full"
       } overflow-y-auto md:-translate-x-0`}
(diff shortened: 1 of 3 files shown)


components/transferForm.tsx
```diff
@@ -308,6 +322,9 @@ function ExecuteContractForm(
           setField={(lambda: string, metadata: string) => {
             props.setField(lambda, metadata);
           }}
+          // onChange={shape => {
```
rueshyna, line 322: remove it?

components/Sidebar.tsx
```diff
@@ -169,7 +169,7 @@ const Sidebar = ({
   return (
     <aside
       className={`fixed left-0 bottom-0 ${
-        state.hasBanner ? "top-32" : "top-20"
+        state.hasBanner ? "top-32" : "top-16"
```
rueshyna, line 172: I think `top-20` looks nicer

this is `top-16`:
image

rueshyna (commented):
The others look good to me :100:

components/Sidebar.tsx
JulesGuesnon: What? That's quite weird, on my screen there was a gap, that's why I had to reduce it, I'll double check

JulesGuesnon: Not sure what happened, I just reversed the change and it was ok

PR #74, opened 2023-05-03
[Refactor] Mutez to Tez

JulesGuesnon:
# Description

This PR remove all mutez references and replace everything by Tez so it's easier for users.

There's one important question that needs to be answer tho: should we write `xtz` or `tez` everywhere?

There's also some small UI fixes in it

### DIFF
--- a/utils/tez.ts
+++ b/utils/tez.ts
@@ -1,0 +1,9 @@
+import BN from "bignumber.js";
+
+const MUTEZ_TEZ_RATIO = Math.pow(10, 6);
+
+export const mutezToTez = (mutez: number): number =>
+  new BN(mutez).div(MUTEZ_TEZ_RATIO).toNumber();
+
+export const tezToMutez = (tez: number): number =>
+  new BN(tez).multipliedBy(MUTEZ_TEZ_RATIO).toNumber();
--- a/versioned/interface.ts
+++ b/versioned/interface.ts
@@ -197,7 +197,7 @@ abstract class Versioned {
         fields: [
           {
             field: "amount",
-            label: "Amount",
+            label: "Amount (Tez)",
             path: ".amount",
             placeholder: "1",
             validate: (x: string) => {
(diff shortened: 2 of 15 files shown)


rueshyna:
> There's one important question that needs to be answer tho: should we write xtz or tez everywhere?

There are 3 symbol, actually.
- Tzkt uses `ꜩ` : https://tzkt.io/tz3gLTu4Yxj8tPAcriQVUdxv6BY9QyvzU1az/operations/
- Temple wallet uses `TEZ`
- TzStats uses `XTZ`

I think prefer `TEZ`. with `TEZ`, if want to present `MUTEZ` one day, it seems to be easier. Maybe it won't be case. 
The second preference of mine is `ꜩ`
The button line is we choice only one and always use it.

utils/tez.ts
```diff
@@ -0,0 +1,9 @@
+import BN from "bignumber.js";
+
+const MUTEZ_TEZ_RATIO = Math.pow(10, 6);
+
+export const mutezToTez = (mutez: number): number =>
```
rueshyna, line 5: Can you add some unit test for these two functions?
and make CI run test when build?

versioned/interface.ts
```diff
@@ -197,7 +197,7 @@ abstract class Versioned {
         fields: [
           {
             field: "amount",
-            label: "Amount in Mutez",
```
rueshyna, line 200: I think adding unit is need so maybe like `Amount(Tez)`?

rueshyna (commented):
The others looks good to me

versioned/interface.ts
rueshyna: so as the others

JulesGuesnon:
@rueshyna I added the test inside the dockerfile, so everything will stop if tests fails

rueshyna (approved):
Please change xtz to Tez before merge
```
> ag xtz 
components/topUpForm.tsx
100:                  Failed to transfer xtz
108:            Amount of xtz to transfer

components/Sidebar.tsx
71:        <p className="text-lg">{formattedBalance.toString()}xtz</p>
```

PR #79, opened 2023-05-05
[Fix] Allowing floats in amounts

JulesGuesnon:
# Description

Because previous amounts were in Mutez, only ints where allowed, so now floats as well are valid inputs


rueshyna (approved):
The change looks good to me.
There is one thing, I missed on previous PR. 

- Amount in "FA2 Transfer" shouldn't have unit
- Amount in "Contract Execution" should have unit.

image

Please help correct it before merge.

PR #89, opened 2023-05-16
[Feature] Multiple FA2 tokens transfer

JulesGuesnon:
# Description
This PR adds allow to transfer multiple fa2 tokens from the same contract at the same time:
- You can new fields
- New fields should only search tokens from the same contract than the first
- Once a token is selected it can't appear anymore in the search
- if first token change, it reset all the tokens after

### DIFF
--- a/components/FA2Transfer.tsx
+++ b/components/FA2Transfer.tsx
@@ -1,19 +1,14 @@
+import { PlusIcon } from "@radix-ui/react-icons";
 import { validateAddress, ValidationResult } from "@taquito/utils";
 import { ErrorMessage, Field, FieldInputProps } from "formik";
 import { useCallback, useContext, useEffect, useRef, useState } from "react";
+import { v4 as uuidV4 } from "uuid";
 import { API_URL, THUMBNAIL_URL } from "../context/config";
 import { AppStateContext } from "../context/state";
 import { debounce } from "../utils/timeout";
 import Alias from "./Alias";
 import Select from "./Select";
 import renderError from "./formUtils";
-
-type props = {
-  index: number;
-  remove: (index: number) => void;
-  setFieldValue: (name: string, value: fa2Token | string) => void;
-  getFieldProps: (name: string) => FieldInputProps<fa2Token | undefined>;
-};
 
 type fa2Token = {
   id: number;
@@ -41,7 +36,7 @@ type option = {
   value: string;
   label: string;
   tokenId: string;
-  thumbnailHash: string | undefined;
+  image: string | undefined;
   contractAddress: string;
   token: fa2Token;
 };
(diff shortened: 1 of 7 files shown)


components/FA2Transfer.tsx
```diff
@@ -110,13 +123,15 @@ const FA2Transfer = ({
...
+          return Promise.resolve(
+            v.filter(token => !toExclude.includes(token.id))
+          );
         }),
     [state.currentContract]
   );
 
   useEffect(() => {
-    const value = getFieldProps(makeName("token")).value;
+    const value = getFieldProps(makeName("token")).value as fa2Token;
```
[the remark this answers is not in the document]
JulesGuesnon: Basically it can't happen as `formValue[]` is valid for this name: `transfers.${proposalIndex}.values` and not for `transfers.${proposalIndex}.values.${index}.token`. But this is the only way to avoid an `any` here: `FieldInputProps<fa2Token | formValue[] | undefined>`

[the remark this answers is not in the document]
rueshyna: but does it mean it could never be `undefined` as well? we tell compile `value` can only be `fa2Token`. Therefore, we don't need line #139 ` if (!value) return;`

so if in ocaml/haskell, we would do pattern match for each case, like
```
match ... with
| fa2Token -> ...
| formValue[] -> ...
| undefined -> ...
```

to make sure each case is handled properly.  If some cases are not going to happen, throwing error message is a good idea. sometimes, error is not only for users, it's also developers.  as last time, you accidentally remove the input field in contract execution. if there isn't an error throwing, we may end up error report by users.

rueshyna (commented):
The layout looks good! I guess it's last bug:

1. select 2 tokens
image
2. click submit and reject
3. the second is gone.
image

context/fa2.ts
```diff
@@ -6,18 +6,26 @@ type makeFa2MichelsonParam = {
...
     DROP;
-    PUSH address "${fa2Address}";
+    PUSH address "${params[0].fa2Address}";
     CONTRACT %transfer (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount))))));
     IF_NONE { PUSH string "contract dosen't exist" ; FAILWITH } { } ;
     PUSH mutez 0 ;
-    PUSH (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))) {Pair "${walletAddress}" {Pair "${targetAddress}" (Pair ${tokenId} ${amount})}} ;
+    PUSH (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))) {Pair "${
+      params[0].walletAddress
+    }"
```
rueshyna, line 20:
```suggestion
    }"  {
```
sorry :pray:  , I made wrong suggestion again ... that was only work for one. this time should be right.

context/fa2.ts
```diff
@@ -6,18 +6,26 @@ type makeFa2MichelsonParam = {
...
     IF_NONE { PUSH string "contract dosen't exist" ; FAILWITH } { } ;
     PUSH mutez 0 ;
-    PUSH (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))) {Pair "${walletAddress}" {Pair "${targetAddress}" (Pair ${tokenId} ${amount})}} ;
+    PUSH (list (pair (address %from_) (list %txs (pair (address %to_) (pair (nat %token_id) (nat %amount)))))) {Pair "${
+      params[0].walletAddress
+    }" 
+      ${params
+        .map(
+          ({ targetAddress, tokenId, amount }) =>
+            `{Pair "${targetAddress}" (Pair ${tokenId} ${amount})} ;`
```
rueshyna, line 24:
```suggestion
            `Pair "${targetAddress}" (Pair ${tokenId} ${amount}) ;`
```

context/fa2.ts
```diff
@@ -6,18 +6,26 @@ type makeFa2MichelsonParam = {
...
+        )
+        .join("\n")}
+      
+    } ;
```
rueshyna, line 28:
```suggestion
    } };
```

rueshyna (commented):
besides suggestion. the others looks good.

when reject and fail to create proposal, it can't submit again.
image

components/FA2Transfer.tsx
```diff
@@ -81,22 +92,43 @@ const FA2Transfer = ({
...
+    }
   };
 
   const fetchTokens = useCallback(
     (value: string, offset: number) =>
       fetch(
-        `${API_URL}/v1/tokens/balances?account=${state.currentContract}&offset=${offset}&limit=${FETCH_COUNT}&token.metadata.name.as=*${value}*&balance.ne=0&sort.desc=lastTime&token.standard.eq=fa2`
+        `${API_URL}/v1/tokens/balances?account=${
+          // state.currentContract
+          "tz1VSUr8wwNhLAzempoch5d6hLRiTh8Cjcjb"
```
JulesGuesnon, line 128: After you tested, can you just delete the address and uncomment the line before merging it?

rueshyna (approved):
:100:

rueshyna (approved):
:100:

PR #95, opened 2023-06-12
[Fix] Warning when adding or removing signers

JulesGuesnon:
# Description

This PR add a warning before signing the proposal to make user is aware that this change will apply on all the proposals

### DIFF
--- a/components/proposalSignForm.tsx
+++ b/components/proposalSignForm.tsx
@@ -201,6 +201,14 @@ function ProposalSignForm({
                 />
               ))}
             </div>
+            {!!proposal.ui.content.find(
+              v => "addOwners" in v || "removeOwners" in v
+            ) && (
+              <span className="mt-2 text-xs font-light text-yellow-500">
+                This proposal will update the signers for all the active
+                proposals
+              </span>
+            )}
           </section>
           <p className="mt-8 text-lg font-medium text-white">
             Action:{" "}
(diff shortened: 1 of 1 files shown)


rueshyna (commented):
we also need to put warning message for the proposals of `changeThreshold` and `adjustEffectivePeroid`

components/proposalSignForm.tsx
```diff
@@ -202,7 +202,11 @@ function ProposalSignForm({
               ))}
             </div>
             {!!proposal.ui.content.find(
-              v => "addOwners" in v || "removeOwners" in v
+              v =>
+                "addOwners" in v ||
+                "removeOwners" in v ||
+                "changeThreshold" in v ||
+                "adjustExpirationPeriod" in v
```
rueshyna, line 209: the message doesn't show.
image

components/proposalSignForm.tsx
```diff
@@ -202,7 +202,11 @@ function ProposalSignForm({
...
             ) && (
               <span className="mt-2 text-xs font-light text-yellow-500">
                 This proposal will update the signers for all the active
```
rueshyna, line 212:
```suggestion
                This proposal will update the setting for all the active
```
in case, it's not a case of adding/removing owners.

components/proposalSignForm.tsx
```diff
@@ -61,6 +61,8 @@ function ProposalSignForm({
     );
   }
 
+  console.log(proposal.ui.content);
```
rueshyna, line 64: remember to remove this before merge this PR.

rueshyna (approved):

PR #96, opened 2023-06-12
[fix] decodeb58 certains type for contract execution

rueshyna:
before:
image

after:
image

limit:
it won't decode `lamdba`. (leave it to future work)

### DIFF
--- a/context/parseLambda.ts
+++ b/context/parseLambda.ts
@@ -4,6 +4,7 @@ import {
   validateAddress,
   ValidationResult,
 } from "@taquito/utils";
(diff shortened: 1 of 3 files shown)


rueshyna:
Maybe `function decodeB58(type: Expr, data: Expr): Expr` should be `function decodeB58(type: Expr, data: Expr): void` ?
@JulesGuesnon

JulesGuesnon (commented):
LGTM, only have to do the fix you mentionned

JulesGuesnon (approved):

PR #101, opened 2023-06-15
[Fix] Proposal resolution threshold

JulesGuesnon:
# Description

This PR fix the detection of if a proposal should be resolved or not


rueshyna (approved):
:+1: 
Let's roll it out!

PR #104, opened 2023-06-19
[Feat] Handling FA2 and FA1.2 in history

JulesGuesnon:
# Description
This PR detects when FA1.2 or FA2 tokens are sent to the wallet

# Overview

image

image

### DIFF
--- a/components/HistoryToken.tsx
+++ b/components/HistoryToken.tsx
@@ -1,0 +1,93 @@
+import { TriangleDownIcon } from "@radix-ui/react-icons";
+import BigNumber from "bignumber.js";
+import { useMemo, useState } from "react";
+import { TransferType } from "../types/display";
+import Alias from "./Alias";
+import { fa1_2Token } from "./FA1_2";
+import { fa2Token } from "./FA2Transfer";
+
+type props = {
+  transferType: TransferType;
+  token: {
+    timestamp: string;
+    from: {
+      address: string;
+    };
+    amount: string;
+    token: fa1_2Token["token"] | fa2Token["token"];
+  };
+};
+
+const HistoryToken = ({ transferType, token }: props) => {
+  const [isOpen, setIsOpen] = useState(false);
+
+  const { tzDate, tzHours, tzMinutes } = useMemo(
+    () => ({
+      tzDate: new Date(token.timestamp).toLocaleDateString(),
+      tzHours: new Date(token.timestamp).getHours().toString().padStart(2, "0"),
+      tzMinutes: new Date(token.timestamp)
+        .getMinutes()
+        .toString()
+        .padStart(2, "0"),
+    }),
+    [token.timestamp]
+  );
+
+  return (
+    <div
+      key={token.timestamp}
+      className={`${
+        isOpen ? "h-full" : "h-16"
+      } w-full overflow-hidden rounded bg-zinc-800 text-white`}
+    >
+      <button
+        className="grid h-16 w-full grid-cols-4 items-center gap-8 border-b border-zinc-900 px-6 py-4 lg:grid-cols-5"
+        onClick={() => setIsOpen(!isOpen)}
+      >
+        <span className="justify-self-start font-bold md:ml-11">
+          <span className="hidden md:block">
+            Received {transferType === TransferType.FA2 ? "FA2" : "FA1.2"}
+          </span>
+          <span className="md:hidden">Received</span>
+        </span>
+        <span className="text-center font-light text-zinc-300 md:min-w-[7rem] md:text-left">
+          <span className="hidden md:inline">From:</span>{" "}
+          <Alias address={token.from.address} />
+        </span>
+        <span className="truncate font-light text-zinc-300 md:min-w-[7rem]">
+          <span className="hidden md:inline">Amount:</span>{" "}
+          {BigNumber(token.amount)
+            .div(BigNumber(10).pow(token.token.metadata.decimals))
+            .toNumber()}{" "}
+          {token.token.metadata.symbol}
+        </span>
+        <span className="hidden justify-self-end lg:block">
+          {tzDate} - {`${tzHours}:${tzMinutes}`}
+        </span>
+        <div className="justify-self-end">
+          <TriangleDownIcon
+            className={`${isOpen ? "rotate-180" : ""} h-8 w-8`}
+          />
+        </div>
+      </button>
+      <div className="px-4 py-4 md:pl-16">
+        <ul className="font-light">
+          <li>
+            <span className="text-zinc-400">Full name:</span>{" "}
+            {token.token.metadata.name}
+          </li>
+          <li>
+            <span className="text-zinc-400">Token id:</span>{" "}
+            {token.token.tokenId}
+          </li>
+          <li>
+            <span className="text-zinc-400">Contract address:</span>{" "}
+            <Alias address={token.token.contract.address} />
+          </li>
+        </ul>
+      </div>
+    </div>
+  );
+};
+
+export default HistoryToken;
--- a/pages/history.tsx
+++ b/pages/history.tsx
@@ -1,20 +1,32 @@
 import { tzip16 } from "@taquito/tzip16";
 import { validateContractAddress } from "@taquito/utils";
+import BigNumber from "bignumber.js";
 import { useContext, useEffect, useMemo, useState } from "react";
 import Alias from "../components/Alias";
+import HistoryToken from "../components/HistoryToken";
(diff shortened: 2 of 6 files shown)


pages/history.tsx
```diff
@@ -103,8 +125,25 @@ const History = () => {
...
+        )
+        .concat(
+          transfers[1].map(
+            x =>
+              [
+                x.token.standard === "fa2"
+                  ? TransferType.FA2
+                  : x.token.standard === "fa1.2"
+                  ? TransferType.FA1_2
+                  : TransferType.UNKNOWN,
```
aguillon, line 144: I didn't read the rest of the code (TL;DR, sorry sorry) but how do you decide the standard of a token in TzSafe?

components/HistoryToken.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <span className="text-zinc-400">Full name:</span>{" "}
+            {token.token.metadata.name}
+          </li>
+          <li>
+            <span className="text-zinc-400">Token id:</span>{" "}
+            {token.token.tokenId}
+          </li>
+          <li>
+            <span className="text-zinc-400">Contract address:</span>{" "}
+            <Alias address={token.token.contract.address} />
```
aguillon, line 85: This appears to shorten the contract's address even when we have enough room on the page (e.g., KT1RJ...peGq4). Why?

aguillon (changes requested):
I'm really not into React so I won't comment on your style or code correctness. I trust you about this!

I like that I can just transfer a token to a multisig from my wallet and that it appears in the history with the name I gave at the multisig creation and not my account address. However, in the future we're going to have people managing tokens with various external contracts (e.g., Youves vaults) and all they're going to read is KT1somethingsomething. We should start thinking about extending aliases now, for instance by using [the same aliases as tzkt](https://baking-bad.org/blog/2021/09/19/tzkt-alias-verification-f-a-q/). This isn't urgent, but I'm curious about your take on this.

I don't think the way tokens are displayed during proposals or after executing them is good though:
proposal



I also think you want to show the token's icon somewhere in the proposal or history. I haven't tested with artsy NFTs, but I guess that it only gives you the token ID and contract address, and that won't cut it for most users.

Other bugs: 
- What's the "metadata" field for? Why is amount = 0? I'm testing here with a FA2 token. It seems to work better for FA1.2.
- Let's assume I have 2 FA1.2 tokens. When submitting a proposal, if I try to transfer 1 to one address, and 2 to another address (which can't work because 1 + 2 = 3 > 2) then I can still submit the proposal, and, later, sign it. Obviously I then can't execute it correctly.
- When making a proposal, if I add a "Transfer FA2" item and fill the "address" field first, and _then_ select a token, I get an "Invalid address" message below the address. Removing one character and then putting it back fixes this, but it's still a minor UI bug.  

I can open separate issues for these bugs if you prefer.

rueshyna:
@aguillon 

for alias, I opened the issue: https://github.com/marigold-dev/tzsafe-ui/issues/106

@JulesGuesnon If the followings are easy to fix, please process. if not, I'm ok with creating an issue for tracking except the last one.

> I also think you want to show the token's icon somewhere in the proposal or history. I haven't tested with artsy NFTs, but I guess that it only gives you the token ID and contract address, and that won't cut it for most users.

I agree that having icon would be nicer.

> - What's the "metadata" field for? Why is amount = 0? I'm testing here with a FA2 token. It seems to work better for FA1.2.

The metadata is for advance users who create their own customized lambda. If users create proposals by tzsafe-ui, metadata is always empty now. the amount of FA2 is showing in `Params/Tokens`. if users transfer multiple FA2 tokens, it's a bit hard to show in the current UI. 
I do agree we should have nicer presentation for regular users. The current proposal presentation is still toward developers. I think this one is a bit out of scope of this PR. 

> - Let's assume I have 2 FA1.2 tokens. When submitting a proposal, if I try to transfer 1 to one address, and 2 to another address (which can't work because 1 + 2 = 3 > 2) then I can still submit the proposal, and, later, sign it. Obviously I then can't execute it correctly.

When the time users create a proposal, it doesn't really execute. Therefore, users have time to prepare funding more FA1.2 token before resolving the proposals. I think showing warnings for users would be good, not restrict them to create these kinds of proposals. It's the same for XTZ transfer.

> - When making a proposal, if I add a "Transfer FA2" item and fill the "address" field first, and then select a token, I get an "Invalid address" message below the address. Removing one character and then putting it back fixes this, but it's still a minor UI bug.

It's a bug. should be fixed.

components/HistoryToken.tsx
rueshyna: wouldn't it be nice to read :o ? the block explorer also do the same practice.

pages/history.tsx
```diff
@@ -103,8 +125,25 @@ const History = () => {
...
+                { ui: { timestamp: x.timestamp }, ...x },
+              ] as any
```
rueshyna, line 146: I don't like this `any` but it looks like it has to change a lot..
so ... ok to leave it here.

rueshyna (commented):
In general, it looks good to me

components/HistoryToken.tsx
aguillon: That's true. I guess here I was surprised by all the empty space.

pages/history.tsx
```diff
@@ -167,58 +206,78 @@ const History = () => {
...
-                      proposer={x[1].og.proposer}
-                      resolver={x[1].og.resolver}
-                      walletTokens={walletTokens}
-                    />
-                  );
+                  switch (x[0]) {
+                    case TransferType.MUTEZ:
+                      return (
+                        <div
+                          key={(x[1] as any).timestamp}
```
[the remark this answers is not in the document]
JulesGuesnon: About the `any` in the history, I definitely agree we should get rid of them, but the thing is, there some data spreaded here and there, so it would need a good refactor to remove them all. Tbh I can do it, but as we have other priorities and it's not that much a problem, we can postpone it to later I think. Wdyt?

JulesGuesnon:
@rueshyna 

> I agree that having icon would be nicer.

This is easy to add because I do have the data to do it, but my main problem is basically where to add it. I quickly thought about it, and it doesn't really fit with current history design I think

> It's a bug. should be fixed.

On it

For the rest I think it can be turned into issues

pages/history.tsx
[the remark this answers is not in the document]
rueshyna: Sure, if it takes effort. Close thread.

rueshyna:
> This is easy to add because I do have the data to do it, but my main problem is basically where to add it. I quickly thought about it, and it doesn't really fit with current history design I think

That's what I was thinking of. issue: https://github.com/marigold-dev/tzsafe-ui/issues/108

components/HistoryToken.tsx
```diff
@@ -0,0 +1,93 @@
...
+    timestamp: string;
+    from: {
+      address: string;
+    };
+    amount: string;
+    token: fa1_2Token["token"] | fa2Token["token"];
+  };
+};
+
+const HistoryToken = ({ transferType, token }: props) => {
```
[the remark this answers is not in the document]
rueshyna: minor: I mean also check transferType inside function. so if one day someone accidentally pass other `MUTEZ`,  the  line49 will not treat it as `FA1.2`

rueshyna (approved):
:+1:

JulesGuesnon:
@aguillon 

> I don't think the way tokens are displayed during proposals or after executing them is good though:

I agree with this, there is probably a way more visual solution to come up with as we have icons for FA1.2 and FA2. Maybe rather than a big table that is definitely made for developers, we could show some small cards for each transaction. And each card could have an icon: 
- Token icon if there's some
- Custom icon for Tzsafe related operations (adding user, deleting user, updating proposal duration, etc...)

> [the same aliases as tzkt](https://baking-bad.org/blog/2021/09/19/tzkt-alias-verification-f-a-q/).

I see the point, and I think it would be valuable to have that kind of feature, but I don't really see how we can use it with our current features. I mean, it can probably be used to have a feature of verified address, and provide more informations to the user, but as long as addresses in TzSafe are only strings that you fill into inputs it's quite complicated I think. The only way I can see an integration of this feature for now, is fetching an API to see if the address is verified on tzkt, and provide a link to the user if it exists so he can check by himself

pages/history.tsx
JulesGuesnon: It's provided by tzkt api. which actually led to some weird bug because tzkt was identifying some contract being fa2 when it's wasn't compliant with the spec

PR #107, opened 2023-07-03
[Fix] exact match lambda pattern

rueshyna:
If the lambda pattern does not exactly match the pattern of an FA2 transfer, FA1 transfer/approval, or (un)setting delegate, it will be classified as a lambda execution.

cc: @tj825

### DIFF
--- a/context/parseLambda.ts
+++ b/context/parseLambda.ts
@@ -108,46 +108,102 @@ const rawDataToData = (rawData: Expr, currentParam: param): data => {
   return [];
 };
 
+const parsePattern = <T>(
+  lambda: Expr[],
+  idx: number,
+  instr: string,
+  process: (expr: Prim) => [boolean, T]
+): [boolean, T?] => {
+  const expr = lambda[idx];
+  if (!!expr && "prim" in expr && expr.prim === instr) {
+    const [res, v] = process(expr);
+    return [true && res, v];
+  } else {
+    return [false, undefined];
+  }
+};
+
 const parseDelegate = (
   lambda: Expr[]
 ):
   | [true, LambdaType, { address: string } | undefined]
   | [false, undefined, undefined] => {
-  if (!lambda.find(expr => "prim" in expr && expr.prim === "SET_DELEGATE"))
+  const delegate_instr_size = 4;
+
+  if (lambda.length != delegate_instr_size)
     return [false, undefined, undefined];
 
-  const type = !!lambda.find(
-    expr =>
-      "prim" in expr &&
-      expr.prim === "PUSH" &&
-      // @ts-expect-error
-      expr.args?.[0].prim === "key_hash"
-  )
-    ? LambdaType.DELEGATE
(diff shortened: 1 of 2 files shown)


rueshyna:
if import `KT1AJjYhXrqqWfwsUJPUGAv32GkiKPkehbnr` on ghostnet, the proposal #09 is considered as set_delegate, but it isn't.
after apply this PR, the proposal #09 will be lambda execution.

context/parseLambda.ts
```diff
@@ -4,6 +4,7 @@ import {
   validateAddress,
   ValidationResult,
 } from "@taquito/utils";
+import { parse } from "path";
```
JulesGuesnon, line 7: It looks like an auto-import from the IDE that is unused

context/parseLambda.ts
```diff
@@ -108,46 +109,102 @@ const rawDataToData = (rawData: Expr, currentParam: param): data => {
...
+
+const parseUnDelegate = (
+  lambda: Expr[]
+): [true, LambdaType] | [false, undefined] => {
+  const delegate_instr_size = 3;
+
+  if (lambda.length != delegate_instr_size) return [false, undefined];
+
+  const fail_parse: [boolean, undefined] = [false, undefined];
+  const succ_parse: [boolean, undefined] = [true, undefined];
```
JulesGuesnon, line 188: Just for naming consistency, can you convert it to camel case

context/parseLambda.ts
```diff
@@ -171,110 +228,156 @@ export const parseLambda = (
...
+  // FA1.2 and FA2 is a special case of contract_execution. There number of instrucions is the same as regular contract execution
+  const contract_execution_instr_size = 7;
+  if (lambda.length != contract_execution_instr_size)
+    return [LambdaType.LAMBDA_EXECUTION, undefined];
+
+  const fail_parse: [boolean, undefined] = [false, undefined];
+  const succ_parse: [boolean, undefined] = [true, undefined];
+
+  // parse DROP
+  const [parseDrop] = parsePrimPattern(lambda, 0, "DROP", () => succ_parse);
```
JulesGuesnon, line 258: There's all this variable names: `parseDrop`, `parsePushAddr`, `parseMutez`, etc... Imo the naming is a bit weird because it looks like everything is a function name, which makes it difficult to understand the code. Maybe something like: `couldParseDrop` or something similar can help to make it more clear

JulesGuesnon (commented):
Overall LGTM, only some small details related to code

context/parseLambda.ts
rueshyna: removed

context/parseLambda.ts
rueshyna: modified.

context/parseLambda.ts
rueshyna: modified.

JulesGuesnon (approved):
LGTM

PR #112, opened 2023-07-12
[Fix] Prevent alias from being too long on mobile

JulesGuesnon:
# Description

Fixing this [issue](https://github.com/marigold-dev/tzsafe-issue/issues/32)


rueshyna (approved):
thanks for the fix :+1:

PR #115, opened 2023-08-03
[Feature] support the contract with standard tzip27 and support running in sandbox mode

rueshyna:
*DO NOT MERGE THIS PR*
This version is change quite a lot. I'd like to merge it later and also release all changes we have on ghostnet to mainnet for announcements.

- support the contract with standard tzip27 (version 0.3.0)
  - proposal id is in hex, need to convert to decimal
  - check the diff of `versioned/version0_3_0.ts` by comparing with `versioned/version0_1_1.ts`
     - major change is in function `signProposal` 
-  support running in sandbox mode
  -  run `npm run sandbox`

I think that would be easy to review each commit alone.


deploy/.env.dev
```diff
@@ -0,0 +1,3 @@
+NEXT_PUBLIC_RPC_URL=https://ghostnet.tezos.marigold.dev/
```
JulesGuesnon, line 1: IMO the name of the folder is a bit weird as it's not really related to deploy; but config files. So maybe `config` would be better I think

types/Proposal0_3_0.ts
```diff
@@ -0,0 +1,35 @@
+import { MichelsonMap } from "@taquito/taquito";
```
JulesGuesnon, line 1: Why is it `0.3.0` and not `0.2.0`?

versioned/version0_3_0.ts
```diff
@@ -0,0 +1,530 @@
...
+import { tezToMutez } from "../utils/tez";
+import { promiseWithTimeout } from "../utils/timeout";
+import { ownersForm } from "./forms";
+import { proposals, timeoutAndHash, Versioned } from "./interface";
+
+function convert(x: string): string {
+  return char2Bytes(x);
+}
+
+const proposals_type: MichelsonType = {
```
JulesGuesnon, line 38: Purely for naming consistency, can you rename it in camelCase `proposalsType`

versioned/version0_3_0.ts
```diff
@@ -0,0 +1,530 @@
...
+  }
+  async signProposal(
+    cc: WalletContract,
+    t: TezosToolkit,
+    proposal: number,
+    result: boolean | undefined,
+    resolve: boolean
+  ): Promise<timeoutAndHash> {
+    const proposals: { proposals: BigMapAbstraction } = await cc.storage();
+    const b_proposal = num2PaddedHex(proposal);
```
JulesGuesnon, line 357: Same here

versioned/version0_3_0.ts
```diff
@@ -0,0 +1,530 @@
...
+    const prop: any = await proposals.proposals.get(b_proposal);
+    const batch = t.wallet.batch();
+    if (typeof result != "undefined") {
+      await batch.withContractCall(
+        cc.methods.sign_proposal(b_proposal, prop.contents, result)
+      );
+    }
+    if (resolve) {
+      const schema = new Schema(proposals_type);
+      const proposals_data = schema.Encode(prop.contents);
```
JulesGuesnon, line 367: Same here

versioned/version0_3_0.ts
```diff
@@ -0,0 +1,530 @@
...
+    const transacValue = await promiseWithTimeout(
+      op.transactionOperation(),
+      DEFAULT_TIMEOUT
+    );
+
+    if (transacValue === -1) {
+      return [true, op.opHash];
+    }
+
+    return [false, op.opHash];
```
JulesGuesnon, line 423: I actually wrote some pretty bad code as it comes from me 😅 Would you mind updating it here and on the other versions to
```js
return [transacValue === -1, op.opHash]
```

JulesGuesnon (changes requested):
Overall LGTM, tried it and it worked well!

Only some minor code style feedback

types/Proposal0_3_0.ts
rueshyna: The followings are different contract versions. the interfaces are break and don't compatible with previous one. `0.2.x` supports ticket but we didn't prioritize for UI to support it before.
- 0.1.x: current one used on mainnet.
- 0.2.x: base on 0.1.x and support the ticket 
- 0.3.x: base on 0.1.x and support TZIP-26

versioned/version0_3_0.ts
JulesGuesnon: You updated it in `0_1_1` but not here, can you do it as well ? Once it's done it should be good

rueshyna: I modified in the different function in both file. now it all modified.

JulesGuesnon (approved):

PR #118, opened 2023-08-11
[WIP] [Feat] PoE support

JulesGuesnon:


rueshyna (commented):
haven't finished to read all of code yet.

components/Sidebar.tsx
```diff
@@ -43,6 +43,9 @@ const linkClass = (isActive: boolean, isDisabled: boolean = false) =>
       : "text-zinc-400"
   } hover:text-zinc-100 flex items-center space-x-3`;
 
+const canConnectToDapp = (version: version) =>
+  version === "0.3.0" || version === "0.3.1" || version === "0.3.2";
```
rueshyna, line 47: can we put an abstracted function call `isSupportPoE` in `versioned/interface.ts`? 
then we just call something like `Versioned.isSupportPoE` ? so we don't need to go back here when adding the new version.

pages/beacon.tsx
```diff
@@ -68,14 +70,22 @@ const Beacon = () => {
 
     const data = decodeData((searchParams.get("data") ?? code) as string);
 
+    if (
+      data.appUrl === "http://localhost:3000" ||
+      data.appUrl.includes("tzsafe")
```
rueshyna, line 77: also need to check the `data.name` is "TzSafe" as the first checking condition. 
`http://localhost:3000` should be removed

pages/beacon.tsx
```diff
@@ -68,14 +70,22 @@ const Beacon = () => {
...
+    ) {
+      setError("Sorry you can't pair Tzsafe with itself");
```
rueshyna, line 79:
```suggestion
      setError("Sorry you can't pair TzSafe with itself");
```

pages/beacon.tsx
```diff
@@ -122,7 +132,15 @@ const Beacon = () => {
                       if (!inputRef.current) return;
 
                       try {
-                        decodeData(inputRef.current.value);
+                        const data = decodeData(inputRef.current.value);
+
+                        if (
+                          data.appUrl === "http://localhost:3000" ||
```
rueshyna, line 201: the same

components/Sidebar.tsx
```diff
@@ -491,6 +498,30 @@ const Sidebar = ({
...
...16.212C1.975 15.0207 0.9 13.4667 0.35 11.55C0.266667 11.25 0.296 10.9667 0.438 10.7C0.579333 10.4333 0.8 10.2667 1.1 10.2C1.38333 10.1333 1.63767 10.1957 1.863 10.387C2.08767 10.579 2.24167 10.8167 2.325 11.1C2.75833 12.5667 3.596 13.75 4.838 14.65C6.07933 15.55 7.46667 16 9 16C10.95 16 12.604 15.3207 13.962 13.962C15.3207 12.604 16 10.95 16 9C16 7.05 15.3207 5.39567 13.962 4.037C12.604 2.679 10.95 2 9 2C7.85 2 6.775 2.26667 5.775 2.8C4.775 3.33333 3.93333 4.06667 3.25 5H5C5.28333 5 5.521 5.09567 5.713 5.287C5.90433 5.479 6 5.71667 6 6C6 6.28333 5.90433 6.52067 5.713 6.712C5.521 6.904 5.28333 7 5 7H1C0.716667 7 0.479333 6.904 0.288 6.712C0.0960001 6.52067 0 6.28333 0 6V2C0 1.71667 0.0960001 1.479 0.288 1.287C0.479333 1.09567 0.716667 1 1 1C1.28333 1 1.521 1.09567 1.713 1.287C1.90433 1.479 2 1.71667 2 2V3.35C2.85 2.28333 3.88767 1.45833 5.113 0.875C6.33767 0.291667 7.63333 0 9 0C10.25 0 11.421 0.237333 12.513 0.712C13.6043 1.18733 14.5543 1.829 15.363 2.637C16.171 3.44567 16.8127 4.39567 17.288 5.487C17.7627 6.579 18 7.75 18 9C18 10.25 17.7627 11.4207 17.288 12.512C16.8127 13.604 16.171 14.554 15.363 15.362C14.5543 16.1707 13.6043 16.8127 12.513 17.288C11.421 17.7627 10.25 18 9 18Z"
```
rueshyna, line 518: need another icon.

components/transferForm.tsx
```diff
@@ -482,6 +482,8 @@ function TransferForm(
 
         values.transfers.forEach((element, idx) => {
           Object.entries(element.values).forEach(([labl, value]) => {
+            if (!("fields" in element)) return;
```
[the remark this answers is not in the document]
JulesGuesnon: Because some transfers don't have the fields props, so I need to do this check to have the right types

pages/beacon.tsx
JulesGuesnon: I don't think we should check the name. Anyone can put the name he wants, checking the `appUrl` is better imo

rueshyna: but if the dapp's name  is "TzSafe", it's high chance that it's malicious one.
Usually, you don't use the existed name as your product name.

rueshyna (commented):
I haven't read all of them yet, but I leave some comments.
The `bun.lockb` should be removed.

pages/_app.tsx
```diff
@@ -157,6 +177,8 @@ export default function App({ Component, pageProps }: AppProps) {
           const balance = await state?.connection.tz.getBalance(userAddress);
           dispatch({
             type: "login",
+            // TODO: FIX
```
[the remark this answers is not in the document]
JulesGuesnon: Both are here because the version of beacon and taquito's beacon are not the same, so there's a type error. Hopefully we can just ignore it and it works, but we'll need to remove the `@ts-ignore` once taquito will updates

utils/connectWallet.ts
```diff
@@ -18,6 +20,8 @@ export const connectWallet = async (
   let s = await state?.beaconWallet!.client.getActiveAccount();
   dispatch!({
     type: "login",
+    // TODO: Fix
```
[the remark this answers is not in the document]
JulesGuesnon: Same than above

components/PoeModal.tsx
```diff
@@ -0,0 +1,542 @@
...
+                        type: "contract",
+                        values: {
+                          lambda: generateExecuteContractMichelson(version, {
+                            address: detail.destination,
+                            amount: Number(detail.amount),
+                            entrypoint: detail.parameters.entrypoint,
+                            type,
+                            param,
+                          }),
+                          metadata: JSON.stringify({
```
rueshyna, line 146: we don't need this. metadata should be `null`

components/PoeModal.tsx
```diff
@@ -0,0 +1,537 @@
...
+                            amount: Number(detail.amount),
+                            entrypoint: detail.parameters.entrypoint,
+                            type,
+                            param,
+                          }),
+                          metadata: null,
+                        },
+                      };
+                    } catch (e) {
+                      console.log("Error while converting contract call", e);
```
rueshyna, line 150: - the error message should in the UI.
- also need to handle the error by sending/rejecting response back to client
image

components/PoeModal.tsx
```diff
@@ -0,0 +1,537 @@
...
+    if (!state.p2pClient) return;
+
+    const challengeCb = setMessage;
+    const transactionCb = async (message: OperationRequestOutput) => {
+      setCurrentMetadata([message.id, message.appMetadata]);
+
+      setTransfers(
+        (
+          await Promise.all(
+            message.operationDetails.map(async detail => {
```
rueshyna, line 98: if the array is empty, should also reject/response error to client.

rueshyna (changes requested):

components/PoeModal.tsx
rueshyna: another suggestion:
the pop window should also show which tzsafe wallet is going to created a proposal.
If users import multiple and connected isn't the current one, it will confusing the users.

rueshyna (changes requested):

context/P2PClient.ts
```diff
@@ -0,0 +1,170 @@
...
+          type: BeaconMessageType.Error,
+          errorType: BeaconErrorType.UNKNOWN_ERROR,
+          id: message.id,
+          senderId: await this.beaconId,
+        });
+        console.warn("Unknown message type", message);
+    }
+  };
+
+  async abortRequest(id: string) {
```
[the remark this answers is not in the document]
JulesGuesnon: The abort error can also come from the wallet, but I agree I was using it too much

rueshyna (changes requested):

components/PoeModal.tsx
```diff
@@ -0,0 +1,659 @@
...
+                        BeaconErrorType.TRANSACTION_INVALID_ERROR
+                      );
+                      return undefined;
+                    }
+                  } else
+                    return {
+                      type: "transfer",
+                      values: {
+                        to: detail.destination,
+                        amount: detail.amount,
```
rueshyna, line 217: Need to check if this value is valid.
Sent error back to client for the invalid's case.

components/PoeModal.tsx
```diff
@@ -0,0 +1,659 @@
...
+                          newline: "",
+                        }
+                      );
+
+                      return {
+                        type: "contract",
+                        values: {
+                          lambda: generateExecuteContractMichelson(version, {
+                            address: detail.destination,
+                            amount: Number(detail.amount),
```
rueshyna, line 190: need to check and throw error.
if `detail.amount` is 'a', UI shows `-`. 
When click confirm, UI will show user "Failed to create the transaction. Please try again later". 
It's incorrect message.

components/PoeModal.tsx
```diff
@@ -0,0 +1,659 @@
...
+                        },
+                      }
+                    : undefined;
+
+                default:
+                  return undefined;
+              }
+            })
+          )
+        ).filter(v => !!v) as transfer[]
```
rueshyna, line 242: before `filter` should check if there is `undefined`. If there is, send error back to client.

components/PoeModal.tsx
```diff
@@ -0,0 +1,659 @@
...
+                      },
+                    };
+
+                case TezosOperationType.DELEGATION:
+                  return !!detail.delegate
+                    ? {
+                        type: "lambda",
+                        values: {
+                          lambda: generateDelegateMichelson(version, {
+                            bakerAddress: detail.delegate,
```
rueshyna, line 228: check address is valid address at least.
(nice to check if it's backer.)

rueshyna: image

rueshyna (changes requested):

components/PoeModal.tsx
```diff
@@ -0,0 +1,659 @@
...
+
+              return (
+                <>
+                  <div className="col-span-2 flex w-full flex-col items-center justify-center">
+                    <div className="mb-2 mt-4 self-start text-2xl font-medium text-white">
+                      Incoming action{(transfers?.length ?? 0) > 1 ? "s" : ""}{" "}
+                      from {currentMetadata?.[1].name}
+                    </div>
+                    <p className="self-start text-sm text-zinc-400">
+                      <Alias address={address ?? ""} /> will create the proposal
```
rueshyna, line 439: I'm thinking that put a warning message when `address` isn't equally to current address to reduce confusing ?

components/PoeModal.tsx
```diff
@@ -0,0 +1,694 @@
...
+                  },
+                };
+              }
+
+            case TezosOperationType.DELEGATION:
+              if (
+                validateAddress(detail.delegate ?? "") !==
+                ValidationResult.VALID
+              ) {
+                state.p2pClient?.sendError(
```
rueshyna, line 241: 1. after sendError, should stop continuing to #248
2. runtime error
```
Unhandled Runtime Error
Error: No matching request found!

Source
context/P2PClient.ts (149:16) @ respond

  147 | 
  148 | async sendError(id: string, reason: string, errorType: BeaconErrorType) {
> 149 |   return this.respond({
```

components/PoeModal.tsx
```diff
@@ -130,117 +131,151 @@ const PoeModal = () => {
...
+                    errorMessage,
+                    BeaconErrorType.TRANSACTION_INVALID_ERROR
+                  );
+                  return undefined;
+                }
+              } else {
+                if (
+                  validateAddress(detail.destination) !==
+                    ValidationResult.VALID ||
+                  isNaN(Number(detail.amount))
```
rueshyna, line 216: and  `<0`

PR #123, opened 2023-08-25
[Refactor] New router

JulesGuesnon:
# Description

This PR refactor the current router to have something like: `/<wallet-address>/proposals`. This new router allow to share the wallet to other people, even if they're unauthenticated


components/signersForm.tsx
```diff
@@ -100,12 +100,14 @@ const SignersForm: FC<{
   const duration = useMemo(() => {
     if (
       ["0.0.6", "0.0.8", "0.0.9", "unknown version"].includes(
```
rueshyna, line 102: the latest version is `0.1.1`. should we add it? is it possible that we do something here so when the next time adding the new version, we won't forget here?

context/config.ts
```diff
@@ -1,12 +1,12 @@
 import { NetworkType } from "@airgap/beacon-sdk";
 
 export const RPC_URL =
-  process.env.NEXT_PUBLIC_RPC_URL ?? "https://ghostnet.tezos.marigold.dev/";
+  process.env.NEXT_PUBLIC_RPC_URL ?? "https://ghostnet.ecadinfra.com/";
```
rueshyna, line 4: I think we should change back to the marigold's rpc

pages/[walletAddress]/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  return {
+    redirect: {
+      destination: `/${params.walletAddress}/proposals`,
+      permanent: true,
+    },
+  };
+};
+
+export default function WalletAddressIndex() {
```
Comment (rueshyna) [inline pages/[walletAddress]/index.tsx:19]: what's this function for?

context/config.ts
JulesGuesnon: I updated it when marigold was down, and I forgot to switch it back

pages/[walletAddress]/index.tsx
Comment (JulesGuesnon) [reply pages/[walletAddress]/index.tsx:19]: I have no choice than exporting a component function as it's a page. And as code is only meant to run server side to redirect, we don't care about returning a content as user is redirected on the server level, so before the page load

components/signersForm.tsx
JulesGuesnon: We don't need to, this filter is here because old contracts didn't have the `effective_period`

pages/[walletAddress]/proposals.tsx
```diff
@@ -0,0 +1,308 @@
...
+    openModal: {
+      state: 0,
+      proposal: [undefined, 0],
+    },
+  });
+
+  const previousRefresherRef = useRef(-1);
+  const [refresher, setRefresher] = useState(0);
+
+  useEffect(() => {
```
Comment (rueshyna) [inline pages/[walletAddress]/proposals.tsx:91]: currently, if you have an imported wallet and you use `/<another wallet>/proposals` to visit `another wallet`, the page will show "an imported wallet" first after few second it will generate `another wallet`. suggestion as following:
```
const router = useRouter();
```

```suggestion
  useEffect(() => {
     if (!router.isReady) return;
```

Comment (rueshyna) [reply pages/[walletAddress]/proposals.tsx:91]: fixed, as discussion offline

rueshyna (commented):
for the whole brand new session,  `/`, `/address-book`, `/new-wallet` and `/import-wallet` will keep loading.

rueshyna (approved):
thanks for the effort :+1: 
There is a small bug.  When you have two import wallets,  enter `/` on URL bar will always go the first wallet on the sidebar list. Let's fix in the next PR.

PR #124, opened 2023-09-12
[Feat] Version detection

JulesGuesnon:
# Description
This PR adds a new way to fetch version if tzip16 fail. It'll fetch tzkt api, and use `typeHash` and `codeHash` to determine the contract version.

This PR also fixes the last small issue mentioned in #123

### DIFF
--- a/context/metadata.ts
+++ b/context/metadata.ts
@@ -17,11 +17,17 @@ const dispatch: { [key: string]: version } = {
   "0.1.1": "0.1.1",
 };
 
-const VERSION_HASH: { [k: string]: version } = {
-  "-357299388-2016479992": "0.0.10",
-  "-483287042793087855": "0.0.11",
-  "-483287042-426350137": "0.1.1",
+// Those values are from tzkt api: /v1/contracts
+type typeHash = string;
+type codeHash = string;
+
+// Before 0.0.10, the version is stored on the contract so tzip16 won't failed to retrieve it
+const VERSION_HASH: { [k: `${typeHash}:${codeHash}`]: version } = {
+  "-357299388:-2016479992": "0.0.10",
+  "-483287042:793087855": "0.0.11",
+  "-483287042:-426350137": "0.1.1",
 };
+
 async function fetchVersion(
   metadata: ContractAbstraction<ContractProvider> & {
     tzip16(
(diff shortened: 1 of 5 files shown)


context/metadata.ts
```diff
@@ -16,6 +17,11 @@ const dispatch: { [key: string]: version } = {
   "0.1.1": "0.1.1",
 };
 
+const VERSION_HASH: { [k: string]: version } = {
```
rueshyna, line 20: can you add some comments to explain where this hash coming from and why not have version `< 0.0.10` ?

context/metadata.ts
```diff
@@ -26,11 +32,37 @@ async function fetchVersion(
...
+      );
+
+    return dispatch[version] ?? "unknown version";
   } catch {
     return "unknown version";
   }
 }
 export default fetchVersion;
+
+// async function fetchVersion(address: string): Promise<version> {
```
rueshyna, line 57: should be removed

context/metadata.ts
```diff
@@ -26,11 +32,37 @@ async function fetchVersion(
...
+      .getMetadata()
+      .then(metadata => metadata.metadata.version ?? "unknown version")
+      .catch(_ =>
+        fetch(`${API_URL}/v1/contracts?address=${metadata.address}`)
+          .then(r => r.json())
+          .then(
+            ([{ typeHash, codeHash }]: {
+              typeHash: number;
+              codeHash: number;
+            }[]) => VERSION_HASH[`${typeHash}${codeHash}`] ?? "unknown version"
```
rueshyna, line 46: would it be better like? that would be easy to understand.
```suggestion
            }[]) => VERSION_HASH[[${typeHash},${codeHash}]] ?? "unknown version"
```

context/metadata.ts
```diff
@@ -16,6 +17,17 @@ const dispatch: { [key: string]: version } = {
...
 };
 
+// Those values are from tzkt api: /v1/contracts
+type typeHash = string;
+type codeHash = string;
+
+// Before 0.0.10, the version is stored on the contract so tzip16 won't failed to retrieve it
+const VERSION_HASH: { [k: `${typeHash}:${codeHash}`]: version } = {
+  "-357299388:-2016479992": "0.0.10",
+  "-483287042:793087855": "0.0.11",
```
rueshyna, line 27: It's before `0.0.11`, no need for `0.0.10`
I think the `codeHash` of `0.0.11` is `521053333`.

```suggestion
// Before 0.0.11, the version is stored on the contract so tzip16 won't failed to retrieve it
// typeHash and codeHash are provided by tzkt API
const VERSION_HASH: { [k: `${typeHash}:${codeHash}`]: version } = {
  "-483287042:521053333": "0.0.11",
```

rueshyna (commented):
the other looks good to me.

rueshyna (approved):

PR #125, opened 2023-09-19
[Fix] Sidebar link disabled

JulesGuesnon:
# Description
This PR disabled the sidebar links as long as Tzsafe is loading, so it avoids wrong redirection. It also fixes a an issue that redirected you to not imported wallet when navigating to `/`


rueshyna (approved):
:100:

PR #126, opened 2023-09-22
fix: downgrade formik

rueshyna:
the formkit version > 2.4.1 will cause a bug for using contractExecutoin.
The bug triggered by click adding item into array form.

```
Unhandled Runtime Error
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```


JulesGuesnon (approved):
LGTM

PR #131, opened 2023-10-23
fix:  parsing pair data

rueshyna:
when the pair data represents in optimized form, `array`, the `decodeB58` will fail.

 this wallet `/KT1TCn5UfT1y7mrYBbHWrx5BQsuy8mQN2QwC` will result error on ghostnet.

### DIFF
--- a/utils/contractParam.ts
+++ b/utils/contractParam.ts
@@ -688,7 +688,7 @@ function parseContract(
  *
  *  reference: https://tezos.gitlab.io/active/michelson.html#core-data-types-and-notations
  */
-function toRightAssociative(type: Expr): void {
+function toRightAssociativePairType(type: Expr): void {
   if ("prim" in type && type.prim === "pair" && !!type.args) {
     if (type.args.length <= 2) {
       return;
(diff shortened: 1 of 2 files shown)


utils/contractParam.ts
```diff
@@ -846,25 +876,31 @@ function decodeB58(type: Expr, data: Expr): Expr {
...
         } else {
-          throw new Error("Internal: data should be a prim");
+          throw new Error("Internal: data should be a prim or array");
         }
       }
       case "option": {
         if ("prim" in data) {
           if (data.prim === "Some" && !!data.args) {
             let new_data = type.args?.map((v, i) => {
+              //@ts-expect-error
```
rueshyna, line 900: If I don't add this, compiler will scream...?
@JulesGuesnon  do you have an idea?

JulesGuesnon: It because your code is in a callback. As the scope is different, `data` might change, and typescript assumes the checks are only valid for a given scope. So basically, you have to do again the `Prim` check if you want to avoid `//@ts-expect-error`

JulesGuesnon (approved):
Imo it would be better to get rid of `@ts-expect-error` before merging, but as it's checked and it's mostly some typescript madness, I don't think it's mendatory

utils/contractParam.ts
rueshyna: removed it.

PR #132, opened 2023-10-30
fix: fix decoding

rueshyna:
bug: using wrong decoding to decode key_hash
change:
  - fix decoding
  - using real test data instead
  - correct the description of test case and add more.


JulesGuesnon (approved):
LGTM

PR #133, opened 2023-10-31
[Refactor] Ipfs

JulesGuesnon:
# Description
Marigold's IPFS node is getting deprecated, so we replace it by Piniata. As all the changes are on the `ipfs-proxy`, there's not that much to change in tzsafe.
The main change is that now we rely first on the contract hash to get the version, and in a second time on the metadata if it failed.

⚠️ Before merging, we need to wait for `ipfs-proxy` to be updated by infra


rueshyna (approved):
:100:

PR #134, opened 2023-11-02
support contract 0.3.2 and reduce fetch records on the history page

rueshyna:
- support contract 0.3.2 
  - in the future, if we should spend time to enhance the way to support new version of contract. Now, it becomes a bit painful. the version change is everywhere.. polymorphism is our good friend.
- reduce fetch records on the history page to enhance loading performance

### DIFF
--- a/versioned/interface.ts
+++ b/versioned/interface.ts
@@ -4,6 +4,7 @@ import { validateAddress, ValidationResult } from "@taquito/utils";
 import { BigNumber } from "bignumber.js";
 import { TZKT_API_URL } from "../context/config";
 import { proofOfEventSchema as proposalSchema_0_3_1 } from "../types/Proposal0_3_1";
+import { proofOfEventSchema as proposalSchema_0_3_2 } from "../types/Proposal0_3_2";
 import { contractStorage } from "../types/app";
 import { proposal, version } from "../types/display";
 import { ownersForm } from "./forms";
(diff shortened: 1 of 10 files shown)


versioned/interface.ts
```diff
@@ -50,7 +51,7 @@ abstract class Versioned {
   readonly version: version;
   readonly contractAddress: string;
 
-  public static FETCH_COUNT = 20;
+  public static FETCH_COUNT = 5;
```
JulesGuesnon, line 54: Maybe `10` would be better? I think 5 by 5 may be a bit slow

versioned/interface.ts
```diff
@@ -125,10 +138,23 @@ abstract class Versioned {
...
+          events.map(event => ({
+            key: event.payload.challenge_id,
+            value: proposalSchema_0_3_2.Execute(
+              unpackDataBytes({
+                bytes: event.payload.payload,
+              })
+            ),
+          }))
+        );
+    } else {
```
JulesGuesnon, line 156: I think we can remove this and do a condition on the schema: `(c.version === '0.3.1' ? proposalSchema_0_3_1 : proposalSchema_0_3_2)`

versioned/version0_3_2.ts
```diff
@@ -0,0 +1,400 @@
...
+      balance: balance!.toString() || "0",
+      proposal_map: c.proposals.toString(),
+      proposal_counter: c.proposal_counter.toString(),
+      effective_period: c!.effective_period,
+      threshold: c!.threshold.toNumber()!,
+      owners: c!.owners!,
+      version: "0.3.2",
+    };
+  }
+  private static mapContent(content: content): proposalContent {
```
JulesGuesnon, line 306: `add_or_update_metadata` is not handled at all, even if we don't handle it on the UI side, I think it should be handled here just in case

versioned/interface.ts
rueshyna: I prefer to split them as they are different cases. because:
- otherwise, next time if there is new version, the conditions will be complicated.  
- also you will have to check condition multiple times.

In the future,  I think we refactor "proposalType" to solve this kind of thing.

versioned/interface.ts
rueshyna: I changed to `10`

versioned/version0_3_2.ts
rueshyna: no, it will throw error. as we don't have any unit test here, it's no way to know if you do it right or wrong.  It will be better going with UI.

JulesGuesnon: So should we just ignore it for now?

rueshyna: yes, for now, nothing will go wrong.

JulesGuesnon (approved):

PR #135, opened 2023-11-03
[Fix] Sign form signers

JulesGuesnon:
# Description
The signers were only calculated for local wallets and not fetched ones


rueshyna (commented):
:100:

rueshyna (approved):

PR #136, opened 2023-11-06
 fix: UI only loads the exact contracts it know

rueshyna:
for security reason, from now on,  UI only loads the exact contracts it know.
The change will prevent that malicious contacts act like as tzsafe contract from deceiving users.
according to the TZKT's doc, there is small chance that hash collision may happen.

cc: @tj825


JulesGuesnon (approved):
LGTM

PR #141, opened 2023-11-28
[Fix] Handling beacon connection

JulesGuesnon:
# Description
Beacon was still working, but a better redirection was needed, so here it is

### DIFF
--- a/components/LoginModal.tsx
+++ b/components/LoginModal.tsx
@@ -156,6 +156,10 @@ const LoginModal = ({ data, onEnd }: { data: string; onEnd: () => void }) => {
                   <h1 className="text-center text-lg font-medium">
                     You don{"'"}t have any wallet that is compatible with Tzip27
                   </h1>
+                  <p className="mt-2 text-center text-sm text-zinc-400">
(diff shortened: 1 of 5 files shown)


components/LoginModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+                    </button>
+                  </div>
+                </>
+              );
+
+            case State.REFUSED:
+              return (
+                <>
+                  <h1 className="text-center text-lg font-medium">
+                    You have refused the connection to {parsedData?.name}{" "}
```
rueshyna, line 177:
```suggestion
                    You have refused the connection to {parsedData?.name ?? "DApp"}{" "}
```

components/LoginModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+                          },
+                        });
+                      }}
+                    >
+                      Authorize
+                    </button>
+                  </div>
+                </>
+              );
+            case State.NO_AVAILABLE_WALLET:
```
rueshyna, line 153: If dapp make a connection request to new browser, UI should be in this state as well.

PR #144, opened 2023-12-15
[Feat] Tezos domains support

JulesGuesnon:
# Description

Adding a way to identify Dapps. For now it supports Tezos Domains, and commit, buying a domain and claiming record


components/ProposalCard.tsx
```diff
@@ -84,39 +102,57 @@ const ProposalCard = ({
...
-            minWidth: "7rem",
-          }}
-          title={content
-            .map(v => {
-              return "executeLambda" in v &&
+        <div className="ml-8 flex items-center space-x-4 md:ml-0">
+          {!!dapp && (
+            <div className="relative h-6 w-6 shrink-0">
+              {(() => {
+                switch (dapp) {
```
rueshyna, line 109: Maybe make this switch in `identifyDapp.ts` or component for the config. The goal is centralized the necessary change together. so we don't need to go everywhere for the adding one dapp.

components/ProposalCard.tsx
```diff
@@ -182,36 +218,50 @@ const ProposalCard = ({
...
+                v => v.amount?.includes("*") || v.params?.includes("*")
+              ) && (
+                <div className="mt-2 text-sm text-yellow-500">
+                  * There{"'"}s no decimals
+                </div>
+              )}
+            </>
+          ) : (
+            (() => {
+              switch (dapp) {
```
rueshyna, line 260: the same?

components/RenderProposalContentLambda.tsx
```diff
@@ -434,29 +436,52 @@ export const labelOfProposalContentLambda = (
...
-      ? "Delegate"
-      : type === LambdaType.UNDELEGATE
-      ? "Undelegate"
-      : type === LambdaType.CONTRACT_EXECUTION
-      ? "Execute contract"
-      : type === LambdaType.POE
-      ? "Proof of Event"
-      : "Execute lambda";
+    if (!!dapp) {
+      switch (dapp) {
```
rueshyna, line 448: the same

components/RenderProposalContentMetadata.tsx
```diff
@@ -343,27 +348,50 @@ export const labelOfProposalContentMetadata = (content: proposalContent) => {
...
-      ? "Delegate"
-      : metadata.contract_address ||
-        (!!metadata.meta && metadata.meta?.includes("contract_addr"))
-      ? "Execute contract"
-      : "Execute lambda";
+    if (!!dapp) {
+      const address =
+        metadata.contract_address ?? JSON.parse(metadata.meta).contract_addr;
+
+      switch (dapp) {
```
rueshyna, line 355: the same

components/dapps/tezosDomains/RenderRows.tsx
```diff
@@ -0,0 +1,85 @@
...
+          case tezosDomainsContracts.COMMIT_ADDRESS.mainnet: {
+            if (!("bytes" in micheline)) throw new Error("Wrong params");
+
+            return (
+              <div key={i}>
+                <h3>Commit</h3>
+                <ul className="list-inside list-disc font-light">
+                  <li className="w-1/2 truncate">
+                    Commit hash: {micheline.bytes}
+                  </li>
```
rueshyna, line 32:
```suggestion
                  <li className="w-1/2 truncate">
                    Note: This has no effect unless followed by a buy operation.
                  </li>
```

dapps/tezosDomains.tsx
```diff
@@ -0,0 +1,194 @@
...
+  BUY_ADDRESS,
+  CLAIM_REVERSE_RECORD,
+};
+
+export const tezosDomainsContractsMatcher: contracts = {
+  mainnet: {
+    KT1F7JKNqwaoLzRsMio1MQC7zv3jG9dHcDdJ: true,
+    KT1QHLk1EMUA8BPH3FvRUeUmbTspmAhb7kpd: true,
+    KT1H1MqmUM4aK9i1833EBmYCCEfkbt6ZdSBc: true,
+    KT1TnTr6b2YxSx2xUQ8Vz3MoWy771ta66yGx: true,
```
rueshyna, line 49: `[CLAIM_REVERSE_RECORD.mainnet]: true`?

 so for the other??

dapps/tezosDomains.tsx
```diff
@@ -0,0 +1,195 @@
...
+    label: transactions
+      .flatMap(({ addresses }) => {
+        if (!addresses) return [];
+
+        return [
+          (() => {
+            switch (addresses) {
+              case COMMIT_ADDRESS.mainnet:
+              case COMMIT_ADDRESS.ghostnet:
+                return "Publish intent";
```
rueshyna, line 101: Maybe
```suggestion
                return COMMIT_ADDRESS.name?
```
so in the line 129, it can be
```
              action: COMMIT_ADDRESS.name,
```

dapps/tezosDomains.tsx
JulesGuesnon: The point is to only add const for the contracts that we want to have a custom UI, only having them in const for the keys is not useful I think

rueshyna: but what about line 51?

```
 [BUY_ADDRESS.mainnet]: true,
```

rueshyna (approved):

PR #145, opened 2023-12-18
[feature] display tokens, support metadata and remove beta

rueshyna:
- display NFT on proposal card
  -  bug fix for the fa2 data reading from metadata. the "to" is incorrect.
  - if the first transaction is NFT, tokens will be displayed by default. Otherwise, tokens are closed. 
image

- handle metadata (TZIP16)
  - decode bytes if it's human readable. otherwise, remain the same. 
  - rename matadata for lambda to note to avoid confused users.


components/FA2Display.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const FA2Display: React.FC<{ data: string }> = ({ data }) => {
+  let jsonData;
+
+  try {
+    jsonData = JSON.parse(data);
+  } catch (error) {
+    console.error("Error parsing JSON:", error);
+    return <div>{data}</div>;
+  }
```
JulesGuesnon, line 26: This should be refactored with a useEffect. If we do it this way, it means that on every render we go through this `try/catch` and it's quite heavy

components/FA2Display.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const result = FA2DataCodec.decode(jsonData);
```
JulesGuesnon, line 28: I don't think having data validation is relevant as it's not user input

components/FA2Display.tsx
JulesGuesnon: Also, after a second thought, I don't think this should be here, we should provide to the component the data it needs to render. 
Imo component's job is rendering UI, not doing parsing, so it shouldn't be done here

components/RenderProposalContentLambda.tsx
```diff
@@ -214,18 +230,41 @@ export const contentToData = (
...
+                token.token.tokenId === token_id.toString()
+            );
+            const metadata = token?.token.metadata;
+            let imageUri = metadata?.thumbnailUri;
+
+            if (
+              imageUri === undefined &&
+              metadata !== undefined &&
+              "displayUri" in metadata &&
+              metadata.displayUri !== undefined
```
JulesGuesnon, line 246: `imageUri === undefined` -> `!imageUri`
`metadata !== undefined` -> `metadata` OR `!!metadata`
same for `displayUri`

components/RenderProposalContentLambda.tsx
```diff
@@ -214,18 +230,41 @@ export const contentToData = (
...
+            )
+              imageUri = metadata.displayUri;
+
+            imageUri = imageUri?.includes("ipfs")
+              ? `${THUMBNAIL_URL}/${imageUri.replace("ipfs://", "")}`
+              : imageUri;
```
JulesGuesnon, line 252: It's the 2nd time is this code, refactoring it into a function would be a good idea I think

components/RenderProposalContentMetadata.tsx
```diff
@@ -338,6 +470,8 @@ export const labelOfProposalContentMetadata = (content: proposalContent) => {
     return `Remove signer${content.removeOwners.length > 1 ? "s" : ""}`;
   } else if ("transfer" in content) {
     return `Transfer ${mutezToTez(content.transfer.amount)} Tez`;
+  } else if ("add_or_update_metadata" in content) {
+    return "Updata Metadata(TZIP16)";
```
JulesGuesnon, line 474:
```suggestion
    return "Update Metadata(TZIP16)";
```

components/RenderProposalContentLambda.tsx
```diff
@@ -315,9 +354,22 @@ export const contentToData = (
...
+}) => {
+  let defaultOpen = false;
+  if (
+    isOpenToken &&
+    (data.type == "TransferFA1_2" ||
+      data.type == "ApproveFA1_2" ||
+      data.type == "TransferFA2")
+  )
+    defaultOpen = true;
+  const [hasParam, setHasParam] = useState(defaultOpen);
```
JulesGuesnon, line 372: this code need only to be run 1 time:
```suggestion
  const [hasParam, setHasParam] = useState(() =>  isOpenToken &&
    (data.type == "TransferFA1_2" ||
      data.type == "ApproveFA1_2" ||
      data.type == "TransferFA2"));
```

JulesGuesnon:
Also even if `io-ts` and `fp-ts` look quite cool, adding them only for using it here is not worth it I think. Both libraries are quite heavy:
- `fp-ts`: `4.2mo`
- `io-ts`: `460kb`

I think there's tree shaking, so after compilation we only have the code we're using so it should makes it small, but I'm not sure it's worth it only for 1 component. If the plan is to incrementally refactor all the codebase with it, why not, but I don't think it's the plan for now

components/RenderProposalContentLambda.tsx
rueshyna: moved to util function

components/RenderProposalContentLambda.tsx
rueshyna: The result is different. in the case of `imageUri` is empty string, which is possible. your version work for the empty string. I think, in general, explicit to check the type is good thing unless the writer is very clear what's writing for. I'd say it's hard for writer to always keep an eye such things. I changed for this case.

rueshyna:
> Also even if `io-ts` and `fp-ts` look quite cool, adding them only for using it here is not worth it I think. Both libraries are quite heavy:
> 
> * `fp-ts`: `4.2mo`
> * `io-ts`: `460kb`
> 
> I think there's tree shaking, so after compilation we only have the code we're using so it should makes it small, but I'm not sure it's worth it only for 1 component. If the plan is to incrementally refactor all the codebase with it, why not, but I don't think it's the plan for now

 I did really think about it's good start for refactor, have test case for each of them and start closing to get rid of `any`.
Anyway, I'm trying the powerful tool. so I used the same way to remove this two libs and generate test cases.

package.json
```diff
@@ -38,6 +38,10 @@
     "eslint": "8.28.0",
     "eslint-config-next": "13.0.5",
     "formik": "2.4.1",
+    "fp-ts": "^2.16.1",
+    "io-fs": "^1.0.0",
+    "io-ts": "^2.2.21",
+    "jsdom": "^23.0.1",
```
JulesGuesnon, line 44: This should be moved in `devDependencies` as it's only for tests

components/RenderProposalContentLambda.tsx
```diff
@@ -202,31 +223,40 @@ export const contentToData = (
...
             fa2_address:
               token?.token.contract.address ?? lambda?.contractAddress,
             name: token?.token.metadata.name,
             token_id,
             to: to_,
-            amount: !!token?.token.metadata.decimals
-              ? BigNumber(amount)
-                  .div(BigNumber(10).pow(token?.token.metadata.decimals ?? 0))
-                  .toString()
-              : amount.toString() + "*",
```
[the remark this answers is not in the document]
rueshyna: It's been added in new components. The "*" will be rendered until it display so we can keep amount sill in number in the process. 
FA2: 
https://github.com/marigold-dev/tzsafe-ui/pull/145/files#diff-20ca5b2980ea770677bd4e2284b0a22d91a3a126554d6e4f666388cba844a0f7R51-R53
so as FA1.2

package.json
rueshyna: rmoved.

JulesGuesnon (approved):

PR #147, opened 2024-01-09
fix(Sidebar): Fixed the Tez rouding and added the full amount on hover

JulesGuesnon:
This PR fixes the bug that Tezos Domains found about our rounding


rueshyna (approved):
👍

PR #148, opened 2024-01-12
[Feature] Supporting all tezos domains and objkt endpoints

JulesGuesnon:
# Description
This PR adds custom views for all of the official Tezos domains and Objkt contracts


dapps/tezosDomains.tsx
```diff
@@ -38,42 +39,147 @@ export const CLAIM_REVERSE_RECORD = {
...
+    [COMMIT_ADDRESS.mainnet]: true,
+    [BID.mainnet]: true,
+    [WITHDRAW.mainnet]: true,
+    [SETTLE.mainnet]: true,
     KT1GY5qCWwmESfTv9dgjYyTYs2T5XGDSvRp1: true,
     KT1R4KPQxpFHAkX8MKCFmdoiqTaNSSpnJXPL: true,
     KT1Lu5om8u4ns2VWxcufgQRzjaLLhh3Qvf5B: true,
     KT1VxKQbYBVD8fSkqaewJGigL3tmcLWrsXcu: true,
   },
   ghostnet: {
```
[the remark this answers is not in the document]
rueshyna: another case:  buy the selling domain from others
- example [call](https://ghostnet.tzkt.io/ooCcs6QyqTU9iuADMxbXr7HS57SRbgxrvbcMqyLdSHnFyCYKinW/18577331) 

image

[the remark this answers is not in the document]
rueshyna: When claim the domain you bid, the transfer of FA2 contract will be called.
image

- example [call](https://ghostnet.tzkt.io/oo9qh8hqP2DVbKUzvATfb7ETWqAAsEuR6cgTvfKfxvFzTV8D4Be/18577239)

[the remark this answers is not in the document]
rueshyna: You can cancel the selling offer.

image
- example [call](https://ghostnet.tzkt.io/oo6uRgkDqdjoP3Y9yZVcqtojQioqgBE757gz7bHNQUJGmn59ifj/18577338)

dapps/tezosDomains.tsx
```diff
@@ -217,6 +408,407 @@ export function tezosDomains(transactions: Array<transaction>): CustomView {
...
+                  description: (
+                    <ul className="list-inside list-disc space-y-1 pt-1 font-light">
+                      {data.map(transfer => (
+                        <li>
+                          From: <Alias address={transfer.from_} />
+                          <ul
+                            className="list-inside list-disc"
+                            style={{ marginLeft: "3rem" }}
+                          >
+                            {transfer.txs.map(({ to_, token_id, amount }) => (
```
rueshyna, line 735: As it's the domain is a token, it fits TZIP12, FA2 contract. The following is an example coping from taquito to get domain name back.
I'd like the following change:
- we can show domain instead token ID
- transfer.txs.length should always be 1. If it's not 1, it must some users mass up. because amount will always be 1 for each domain. domain name don't have copy. Therefore, here we can just show "from", "to", "domain name".

so as the other case's involve token ID.
 
```
// import { TezosToolkit, compose } from '@taquito/taquito';
// import { Tzip12Module, tzip12 } from "@taquito/tzip12";
// const Tezos = new TezosToolkit('rpc_url');

Tezos.addExtension(new Tzip12Module());

const contractAddress = "KT1REqKBXwULnmU6RpZxnRBUgcBmESnXhCWs";
const tokenId = 3397;

Tezos.contract.at(contractAddress, tzip12)
.then(contract => {
  println(`Fetching the token metadata for the token ID ${tokenId}...`);
  return contract.tzip12().getTokenMetadata(tokenId);
})
.then (tokenMetadata => {
  println(JSON.stringify(tokenMetadata, null, 2));
})
.catch(error => println(`Error: ${JSON.stringify(error, null, 2)}`));
```
the result:
Fetching the token metadata for the token ID 3397...
{
  "token_id": 3397,
  "decimals": 0,
  "name": "abc.gho",
  "symbol": "DOMAIN"
}

rueshyna (changes requested):

dapps/tezosDomains.tsx
```diff
@@ -38,42 +40,196 @@ export const CLAIM_REVERSE_RECORD = {
...
+  parser.parseMichelineExpression(`(list %update_operators (or
+                         (pair %add_operator (address %owner)
+                                             (pair (address %operator) (nat %token_id)))
+                         (pair %remove_operator (address %owner)
+                                                (pair (address %operator) (nat %token_id)))))`)!
+);
+
+export const TOKEN_CONTRACT = {
+  mainnet: "KT1R4KPQxpFHAkX8MKCFmdoiqTaNSSpnJXPL",
+  ghostnet: "KT1MhAy28Bv8oNb3x71dHy6YP6eiJpDVUiDy",
```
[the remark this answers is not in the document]
JulesGuesnon: No i'm not, and I couldn't find one for ghosnet. Also the most important is to have a contract with the same API so the schemas are matchings, and we can update it easily later

rueshyna (commented):
KT1NPAXF5HtMfsFTASW5gSCtjy7MjJKRHDsa/proposals
There are some error in this contract.
Can you make a switch for mainnet?
Before pushing the change to mainnet, I'd like double confirm with them.
I think this will be common feature if there is plan to integrate with more dapps

dapps/tezosDomains.tsx
```diff
@@ -108,6 +321,37 @@ export function tezosDomains(transactions: Array<transaction>): CustomView {
...
+                return SET_CHILD_RECORD.name;
+              case CHECK_ADDRESS.mainnet:
+              case CHECK_ADDRESS.ghostnet:
+                return CHECK_ADDRESS.name;
+              case OFFER.mainnet:
+              case OFFER.ghostnet:
+                return OFFER.name;
+              case TOKEN_CONTRACT.mainnet:
+              case TOKEN_CONTRACT.ghostnet:
+                return OFFER.name;
```
rueshyna, line 353:
```suggestion
                return TOKEN_CONTRACT.name;
```

dapps/tezosDomains.tsx
```diff
@@ -38,46 +44,253 @@ export const CLAIM_REVERSE_RECORD = {
...
+    node: null,
+  });
+
+  useEffect(() => {
+    promise.then(node => {
+      setPromiseState({ isLoading: false, node });
+    });
+  }, [promise]);
+
+  return <>{node ?? "..."}</>;
```
rueshyna, line 287: How about ?
```suggestion
  return <>{node ?? "loading ..."}</>;
```
Is it easy to add retry? It seems easy to fail?

dapps/tezosDomains.tsx
```diff
@@ -38,46 +44,253 @@ export const CLAIM_REVERSE_RECORD = {
...
+                         (pair %add_operator (address %owner)
+                                             (pair (address %operator) (nat %token_id)))
+                         (pair %remove_operator (address %owner)
+                                                (pair (address %operator) (nat %token_id)))))`)!
+);
+
+export const TOKEN_CONTRACT = {
+  mainnet: "KT1R4KPQxpFHAkX8MKCFmdoiqTaNSSpnJXPL",
+  ghostnet: "KT1REqKBXwULnmU6RpZxnRBUgcBmESnXhCWs",
+  name: "Tezos Domains Token",
```
rueshyna, line 185:
```suggestion
  name: "Tezos Domains NFT",
```

rueshyna (changes requested):

dapps/tezosDomains.tsx
```diff
@@ -183,6 +451,19 @@ export function tezosDomains(transactions: Array<transaction>): CustomView {
...
+                    TzSafe doesn't support the entrypoint:{" "}
+                    {transaction.entrypoints}
+                  </p>
+                ),
+              },
+            ];
+
           const data = claimReverseRecordSchema.Execute(micheline);
 
           const domain = bytes2Char(data.name.Some);
```
rueshyna, line 469: need to check "Some".
To prevent the error:
```
Failed to parse dapp: TypeError: Cannot read properties of null (reading 'Some')
```

dapps/tezosDomains.tsx
JulesGuesnon: Imo, it's not that trivial to add it there

rueshyna (approved):

PR #149, opened 2024-01-16
[feature] support tzsafe 0.3.4

rueshyna:
- support tzsafe contract 0.3.4
- support to select create 0.3.3 or 0.3.4
- add e2e for version 0.1.1, 0.3.3 and 0.3.4
- add update metadata proposal 

Note that proof of event challenge must be through proof_of_event_challenge entrypoint. Therefore, the proof of even proposal will be alone proposal.


components/RenderProposalContentLambda.tsx
```diff
@@ -534,6 +534,10 @@ export const labelOfProposalContentLambda = (
     return `Remove signer${content.removeOwners.length > 1 ? "s" : ""}`;
   } else if ("transfer" in content) {
     return `Transfer ${mutezToTez(content.transfer.amount)} Tez`;
+  } else if ("add_or_update_metadata" in content) {
+    return "Update Metadata(TZIP16)";
+  } else if ("proof_of_event" in content) {
+    return "Message Signing in Proof of event challenge(TZIP27)";
```
JulesGuesnon, line 540: Do we really want to put the Tzip number here?
If so, let's just add a space between the word and the parenthesis

components/create/basic.tsx
```diff
@@ -47,7 +51,44 @@ function Basic() {
...
+                type="radio"
+                name="version"
+                value="0.3.4"
+                className="mr-2"
+              />
+              <div>v0.3.4</div>
+            </label>
+            <div className="ml-2 text-sm text-gray-500">
+              {
+                "This version fully supports for TZIP27. The createion fee is roughly 2.1 tez."
```
JulesGuesnon, line 85:
```suggestion
                "This version fully supports TZIP27. The creation fee is roughly 2.1 tez"
```

components/create/basic.tsx
```diff
@@ -47,7 +51,44 @@ function Basic() {
...
+                name="version"
+                value="0.3.3"
+                className="mr-2 "
+                checked
+              />
+              <div>v0.3.3</div>
+            </label>
+            <div className="ml-2 text-sm text-gray-500">
+              {
+                "This version doesn't support for TZIP27 in processing proof-of-event challenges for message signing. If such functionality is unnecessary, this version is recommended. The creation fee is roughly 1.8 tez, depending on the count of owners."
```
JulesGuesnon, line 69: Not sure how I would say it, but this is unclear imo, we should find a way to say it that is closer to what the user can understand

components/transferForm.tsx
```diff
@@ -16,6 +16,7 @@ import {
   useFormikContext,
 } from "formik";
 import { useRouter } from "next/router";
+import { version } from "os";
```
JulesGuesnon, line 19: To delete

components/transferForm.tsx
```diff
@@ -993,14 +1042,100 @@ function TransferForm(
                                       <ErrorMessage
                                         name={`transfers.${index}.values.${value.field}`}
                                       />
+                                      {!hasTzip27SupportWithPoEChallenge(
+                                        props.contract.version
+                                      ) &&
+                                        renderWarning(
+                                          "This version doesn't support message signing, and this functionality will be removed following submission."
```
JulesGuesnon, line 1049: I don't get what you mean. 
> This version doesn't support message signing, and this functionality will be removed
- in the next version?
- in the following version?

context/fromIpfs.ts
```diff
@@ -0,0 +1,64 @@
...
+        init?: RequestInit | undefined
+      ) => Promise<Response>))
+  | ((arg0: string, arg1: { method: string; body: any; headers: {} }) => any);
+
+if (isNode) {
+  FormDataNode = require("form-data");
+  fetch = require("node-fetch");
+} else {
+  fetch = window.fetch; // Use the browser's fetch
+}
```
JulesGuesnon, line 28: Why do we need to have a specific case for NodeJs? it supposed to run only in the browser

context/generateLambda.ts
```diff
@@ -227,23 +226,3 @@ export function generateUndelegateMichelson(version: version) {
 
   throw new Error("Can't generate for an unknow version");
```
JulesGuesnon, line 227:
```suggestion
  throw new Error("Can't generate for an unknown version");
```

context/version.ts
```diff
@@ -19,6 +26,25 @@ const dispatch: { [key: string]: version } = {
...
+  "0.0.10": undefined,
+  "0.0.11": undefined,
+  //@ts-expect-error There was 2 versions of 0.0.11, so it means 2 codeHash.
+  "0.0.11b": undefined,
+  "0.1.1": [ct0_1_1, meta0_1_1],
+  "0.3.0": undefined,
+  "0.3.1": undefined,
+  "0.3.2": undefined,
+  "0.3.3": [ct0_3_3, meta0_3_3],
+  "0.3.4": [ct0_3_4, meta0_3_4],
```
JulesGuesnon, line 46: Is the tuple the contract code and the metadata?

context/version.ts
```diff
@@ -48,17 +75,18 @@ const VERSION_HASH: { [k: `${typeHash}:${codeHash}`]: version } = {
   [HASHES["0.0.9"]!]: "0.0.9",
   [HASHES["0.0.10"]!]: "0.0.10",
   [HASHES["0.0.11"]!]: "0.0.11",
-  //@ts-ignore
+  //@ts-ignore  This version only exist on ghostnet
```
JulesGuesnon, line 78:
```suggestion
  //@ts-expect-error - This version only exist on ghostnet
```
Just thought about it, but this way we can detect regression

tests/e2eTest/e2eTest.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+vi.mock("@airgap/beacon-sdk", () => ({
+  NetworkType: {
+    MAINNET: "mainnet",
+    GHOSTNET: "ghostnet",
+  },
+}));
+
+test_suit_0_1_1();
+test_suit_0_3_3();
+test_suit_0_3_4();
```
JulesGuesnon, line 15: You way of running the tests works, but it's not really the intended way of doing it. Normally you should:
- Rename `v01.1.1.ts` to `v01.1.1.spec.ts` (and the same for the other files)
- I'm not sure how much it's required, but you have to import `expect`, `it`, `descripe`, from `vitest` ([doc](https://vitest.dev/guide/#writing-tests))

Is there a reason the tests are the way you did it?

tests/e2eTest/v0.3.4.ts
```diff
@@ -0,0 +1,314 @@
...
+const test_suit = () =>
+  describe(`Test version ${version}`, () => {
+    let tezos: TezosToolkit;
+    let addr: string;
+
+    beforeAll(async () => {
+      tezos = new TezosToolkit(rpc);
+      tezos.setProvider({
+        signer: new InMemorySigner(pb),
+      });
```
JulesGuesnon, line 28: It think this could be refactored into a test utils file

versioned/interface.ts
```diff
@@ -678,6 +696,37 @@ abstract class Versioned {
...
+    }
+
+    return {
+      values: {
+        tzip16_metadata: "",
+      },
+      fields: [
+        {
+          field: "tzip16_metadata",
+          label: "Metadata in TZIP16",
```
JulesGuesnon, line 721: Same than before, is putting the Tzip number worth it? 
If it this, then let's do like before and put it between parenthesis

JulesGuesnon (changes requested):

tests/e2eTest/e2eTest.spec.ts
rueshyna: yes, I want them run in order.
If they are in three different files, they will run in parallel.
It's real time to send ops to ghostnet and I make them use the same key. if you run them in parallel, high chance are those op will fail because they are using the same op counter.
It looks like vitest doesn't support for ordering by passing arg or put annotations.

context/fromIpfs.ts
rueshyna: To be able to run e2e test on the docker image. Otherwise, CI keeps complaining and fail.
see the build, like
- https://github.com/marigold-dev/tzsafe-ui/actions/runs/7518831530/job/20466630403

in CI, it will keep complaining about `fromdata`, `blob` and `fetch`.

context/version.ts
rueshyna: yes!

components/transferForm.tsx
rueshyna: It happens when:
1. user is on wallet 0.3.4 and creates a proposal for message signing
2. user switch wallet to 0.1.1 on sidebar. 
So, for wallet 0.1.1, it doesn't support message signing.

versioned/interface.ts
rueshyna: metadata is too common. We also have metadata with lamdba. There are users asking me before so I think it would be clear to add it.

context/fromIpfs.ts
JulesGuesnon: This should be done a test time with mocking as there's no reason to put it at runtime, but it can be fixed later

context/version.ts
JulesGuesnon: Ok then can you change the type: 
`[contractCode: string, metadata: any]`

JulesGuesnon (approved):

PR #164, opened 2024-02-21
Deploy to GitHub Pages

ulrikstrid:


pages/[walletAddress]/index.tsx
Comment (cyrilevos) [inline pages/[walletAddress]/index.tsx:1]: should we find a way to implement the equivalent of whats in getServerSideProps on the client side so that it is compatible with static site exports?

Comment (rueshyna) [reply pages/[walletAddress]/index.tsx:1]: > should we find a way to implement the equivalent of whats in getServerSideProps on the client side
It looks good to me if they are not the same in this case.

is this only question that block to move github page :o ?

Comment (ulrikstrid) [reply pages/[walletAddress]/index.tsx:1]: Fun my testing everything works but someone more familiar with the app should test it.
I think the old (Docker) pipeline is falling with this change because we're building a static site instead of running next.

If we want to keep docker we can change it to leverage nginx or maybe we can make the build understand both usecases.

Comment (rueshyna) [reply pages/[walletAddress]/index.tsx:1]: I had been tested it. It looks good to me. 

Should we really remove dockers? The dockers pass the pipeline 🤔

The failed one is about cloudflare

Comment (rueshyna) [reply pages/[walletAddress]/index.tsx:1]: is there example website already one deployed on github page?

Comment (cyrilevos) [reply pages/[walletAddress]/index.tsx:1]: Then if we're ok with that piece of code being removed yes we can merge that PR. I've been testing it a bit too on my side.
But yes @rueshyna what i can do is to deploy an instance of the Ghostnet WebUI on GHPages first, so that you can test

The thing is that we can only deploy 1 Frontend to GH pages per git repo, that means we would have to either deploy the mainnet or the ghostnet version elsewhere.
I've experimented deploying the ghostnet version on Fly.io, by reusing the Docker image so very minimal change to the CICD and code etc, thus eliminating the need for additional testing. You can see the endpoint here: https://tzsafe-ui-mainnet.fly.dev (yes that's ghostnet ><)

So in the end I suggest we run the Mainnet version on Fly.io, as it provides monitoring and logging features not available on GitHub Pages. We can run the Ghostnet version on GHPages.

Comment (cyrilevos) [reply pages/[walletAddress]/index.tsx:1]: About Cloudflares, i've tested deploying on it complains on static files being too large so i abandoned.

Comment (ulrikstrid) [reply pages/[walletAddress]/index.tsx:1]: The removed code just does some check and then redirects so I don't think it's strictly needed, but I could be wrong

Comment (rueshyna) [reply pages/[walletAddress]/index.tsx:1]: @cyrilevos 
I just play on it. `https://tzsafe-ui-mainnet.fly.dev/`. It looks good to me on `fly.dev` for mainnet.

> The removed code just does some check and then redirects so I don't think it's strictly needed, but I could be wrong
looks good to me.

PR #172, opened 2024-03-27
:bug: redirect to first contract in list instead of last visited

quentin-burg:
Update as we talked this morning.
I'm working on separating contracts imported by the user who imported them on another branch but it's a little more complicated that I thought.


Pilou97 (approved):
LGTM

Issue #106, opened 2023-06-28
[UI][alias] display alias for address


rueshyna:
from @aguillon 
> I like that I can just transfer a token to a multisig from my wallet and that it appears in the history with the name I gave at the multisig creation and not my account address. However, in the future we're going to have people managing tokens with various external contracts (e.g., Youves vaults) and all they're going to read is KT1somethingsomething. We should start thinking about extending aliases now, for instance by using [the same aliases as tzkt](https://baking-bad.org/blog/2021/09/19/tzkt-alias-verification-f-a-q/). This isn't urgent, but I'm curious about your take on this.

There are few tasks about alias in the backlog. We'll go back for this in near future.

oteku:
Rules to apply by priority:

1. Replace KT1 address by TzSafe "Wallet name", if the address is in the wallet list of the connected account
2. Get alias from Tzkt API https://api.tzkt.io/#operation/Accounts_Get
3. Display adress as is

PR #180, opened 2024-04-03
Display alias for addresses

quentin-burg:
Related to https://github.com/marigold-dev/tzsafe-ui/issues/106


utils/getTzktAlias.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from "react";
```
Pilou97, line 1: Can you move this hook to the file `tzktHooks.ts` ? So that everything related to tzkt is in the same file

I never in typescript if we should use `<string| undefined>` or just `string?`

pages/[walletAddress]/history.tsx
```diff
@@ -380,44 +382,10 @@ const History = () => {
             filteredProposals.length > 0 && (
               <div className="space-y-6">
                 {filteredProposals.map(x => {
+                  console.log("x", x, (x[1] as any)?.sender?.address);
```
Comment (Pilou97) [inline pages/[walletAddress]/history.tsx:385]: I think this log is there for debug purpose

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "env-cmd -f ./config/.env.dev next dev",
+    "dev": "env-cmd -f ./config/.env next dev",
```
Pilou97, line 6: why?

components/history/HistoryTransfer.tsx
```diff
@@ -0,0 +1,40 @@
+import { mutezToTez } from "../../utils/tez";
```
Pilou97, line 1: Thanks, splitting component is always a nice to have

components/Sidebar.tsx
```diff
@@ -256,9 +259,11 @@ const Sidebar = ({
                     className="radix-disabled:opacity-50 relative flex select-none items-center rounded-md px-8 py-2 text-sm font-medium text-zinc-300 focus:bg-zinc-800 focus:bg-zinc-900 focus:outline-none"
                   >
                     <Select.ItemText>
-                      <p className="text-xl text-white">
+                      {/* <p className="text-xl text-white">
```
Pilou97, line 262: maybe this comment can be removed?

components/Sidebar.tsx
```diff
@@ -68,7 +69,7 @@ const SelectedItem = ({
         <p className="text-xl text-white">{name}</p>
         <p>{threshold}</p>
       </div>
-      <Copy value={address ?? ""} disabled={disableCopy}>
+      {/* <Copy value={address ?? ""} disabled={disableCopy}>
```
Pilou97, line 72: Does that mean we can't copy anymore the contract address?

components/Alias.tsx
```diff
@@ -25,11 +28,12 @@ const Alias = ({
   const toDisplay = useMemo(
     () =>
       state.aliases[address] === ""
```
Pilou97, line 30: what if  `state.alises[address]` returns null? Maybe it's not possible, feel free to resolve if it's the case

components/Alias.tsx
```diff
@@ -14,6 +15,8 @@ const Alias = ({
   disabled?: boolean;
 }) => {
   const state = useContext(AppStateContext)!;
+  const tzktAlias = useTzktAccountAlias(address);
+  console.log("🚀 ~ tzktAlias:", tzktAlias);
```
Pilou97, line 19: I don't know if logs has to be removed or not, if you think we should keep it we can

Pilou97:
I think there is an issue in the copy pasting of the address (But I am running on Arch Linux, the error can be on my side):

There is still the tooltip "Copy address" but it does not copy it

Pilou97:
Also should we display the alias or the address in the menu entry?

Pilou97 (approved):

PR #183, opened 2024-04-04
Static Pages

Pilou97:
# Context

If you go on "/KT1/dashboard" (or proposals, etc...) you will hit a 404. If not try to reload your page.

It's because this code was deleted in https://github.com/marigold-dev/tzsafe-ui/pull/164

```
import { validateAddress, ValidationResult } from "@taquito/utils";
import Proposals from "./proposals";
import { GetServerSideProps } from "next";


export const getServerSideProps: GetServerSideProps = async ({ params }) => {
export default Proposals;
  if (!params || !params.walletAddress || Array.isArray(params.walletAddress))
    return { notFound: true };

  if (validateAddress(params.walletAddress) !== ValidationResult.VALID)
    return { notFound: true };

  return {
    redirect: {
      destination: `/${params.walletAddress}/dashboard`,
      permanent: true,
    },
  };
};

export default function WalletAddressIndex() {
  return null;
}
```

# Solution

Each time the user is redirect to 404.html (by Github), we check if the page is a valid one and proceed to the redirection. I am not a huge fan of this solution because I am not sure if it's the way to do it with next.js


quentin-burg (approved):
✅

PR #184, opened 2024-04-04
404: directly return when auto redirect from 404

Pilou97:
Don't return the component of the 404 page when there is a redirect


Pilou97:
Thanks for your contribution, I love it :heart:

Issue #187, opened 2024-04-09
[UI][Proposals] Verify account existence when create a new proposal


quentin-burg:
When we create a new proposal, to transfer some tez to an address (implicit account or smart contract) we need to check if this account exists.

PR #188, opened 2024-04-09
:bug: tzkt returns no content if account not exists so don't parse to…

quentin-burg:
… JSON


quentin-burg:
Related to https://github.com/marigold-dev/tzsafe-ui/issues/187 but this PR is a quick fix.

Pilou97 (approved):
Niiice

PR #193, opened 2024-04-18
♻️ Extract wallet to context

quentin-burg:
- Extract BeaconWallet into Context
- Extract TezosToolKit into Context
- Rewrite Context to manage App State and Dispatch function

TODO : Problem with fetching proposals ... To be continued ...


components/LoginModal.tsx
```diff
@@ -51,15 +55,15 @@ const LoginModal = ({ data, onEnd }: { data: string; onEnd: () => void }) => {
   const [currentState, setCurrentState] = useState(() => State.LOADING);
 
   useEffect(() => {
-    if (!state.p2pClient || !state.attemptedInitialLogin) return;
+    if (!state.p2pClient) return;
```
Pilou97, line 55: Why `state.attemptedInitialLogin` is no more necessary?

components/LoginModal.tsx
```diff
@@ -51,15 +55,15 @@ const LoginModal = ({ data, onEnd }: { data: string; onEnd: () => void }) => {
...
 
     try {
       const decoded = decodeData(data);
 
       setParsedData(decoded);
 
       state.p2pClient!.on(Event.PERMISSION_REQUEST, () => {
-        if (state.attemptedInitialLogin && !state.address) {
+        if (!userAddress) {
```
Pilou97, line 63: why `state.attemptedInitialLogin` is no more necessary?

components/LoginModal.tsx
```diff
@@ -77,13 +81,13 @@ const LoginModal = ({ data, onEnd }: { data: string; onEnd: () => void }) => {
       setError((e as Error).message);
       setCurrentState(State.ERROR);
     }
-  }, [data, state.p2pClient, state.attemptedInitialLogin]);
```
Pilou97, line 80: Why `state.attemptedInitialLogin` is no more required?

components/textInputWithComplete.tsx
```diff
@@ -18,7 +18,7 @@ const TextInputWithCompletion: FC<
   }
 > = props => {
   const [field, _, helpers] = useField(props);
-  // const state = useContext(AppStateContext)!;
+  // const state = useAppState();
```
Pilou97, line 21: even the comments are fixed 💪

context/state.tsx
```diff
@@ -376,10 +298,33 @@ function init(): tezosState {
...
+  );
+
+  return (
+    <AppStateContext.Provider value={{ state, dispatch }}>
+      {children}
+    </AppStateContext.Provider>
+  );
+};
+
+const useAppState = () => useContext(AppStateContext).state;
```
Pilou97, line 346: In the old code I see that state is often an optional and accessed by `state?.field`. So we assume that the state is always initialized in your version?

context/wallet.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const initialState: WalletState = {
+  wallet: undefined,
+  userAccount: undefined,
+  userAddress: undefined,
+  userBalance: undefined,
+};
+
+export const WalletContext = createContext<WalletContextType>({
+  state: initialState,
```
Pilou97: why wrapping the initialState in an object ? 🤔 

Not wrapping it would allow to directl access the userAddress like so:

```js
const { userAddress } = useWallet();
```

wdyt?

context/wallet.tsx
```diff
@@ -0,0 +1,151 @@
...
+  connectWallet: () => {},
```
Pilou97, line 54: Can't you directly use `connectWallet(tezos)` defined bellow?

context/wallet.tsx
```diff
@@ -0,0 +1,151 @@
...
+  disconnectWallet: () => {},
```
Pilou97, line 55: same question for `disconnectWallet`

Pilou97 (commented):
First review

context/state.tsx
quentin-burg: Yes because I call `init` function when I create the context. It's his initial value.

context/wallet.tsx
quentin-burg: Yes we can do this. I wrote this because I wanted to split " state logic " to " update functions ".
But it can be merge :)

context/wallet.tsx
quentin-burg: I need `tezos` instance so maybe but I need to check if it's possible

context/wallet.tsx
quentin-burg: Same answer :-p

context/state.tsx
```diff
@@ -162,28 +123,28 @@ const saveState = (state: tezosState) => {
...
-        wallet: action.payload,
-      });
-      state.connection.setSignerProvider(new BeaconSigner(action.payload));
-      return { ...state, beaconWallet: action.payload };
-    }
     case "p2pConnect": {
       return { ...state, p2pClient: action.payload };
     }
     case "addDapp": {
-      if (!state.address) return state;
```
Pilou97, line 179: The address is always defined? I don't know what is `addDapp` but can it be triggered when the user is disconnected without any safe?

dapps/tezosDomains.tsx
```diff
@@ -401,7 +401,7 @@ export function tezosDomains(
                 action: BUY_ADDRESS.name,
                 description: (
                   <p>
-                    TzSafe doesn't support the entrypoint:{" "}
```
Pilou97, line 404: You don't need spaces anymore? I am wondering if we can use template string :thinking:

tests/e2eTest/v0.3.4.ts
```diff
@@ -39,7 +39,7 @@ const test_suit = (setTezosToolkit: (tezos: TezosToolkit) => TezosToolkit) =>
         expect(storage.proposal_counter.isEqualTo(BigNumber(0))).toBe(true);
         expect(storage.owners).toEqual([owner]);
         storage.metadata.get("").then((value: string) => {
-          expect(value).toEqual(char2Bytes(ipfs_file));
+          expect(value).toEqual(stringToBytes(ipfs_file));
```
Pilou97, line 42: what is the difference between `char2Bytes` and `stringToBytes` ? Why do we need to change it?

Pilou97 (commented):
Cool to have update the package.json and airgap

tests/e2eTest/v0.3.4.ts
quentin-burg: `charToBytes` and `bytesToChar` are deprecated => https://taquito.io/typedoc/functions/_taquito_utils.char2Bytes

dapps/tezosDomains.tsx
quentin-burg: Yes you're right :-)

context/state.tsx
quentin-burg: address is no longer defined in the global state but in wallet context.
`addDapp` allows to connect an external dapp with Tzsafe and user have to import a contract to access to this feature, so I believe disconnected user cannot trigger this action

dapps/tezosDomains.tsx
quentin-burg: I updated all of occurrences of this string

